PR #8, opened 2014-10-28
Webmock Helper: Don't break other test suites

lfittl:
## Scope

This changes the mocks in this gem to be less likely to break other project test suites that also use Webmock, by applying the mocks in a rspec before filter instead of directly when you include the file.

In addition this change makes it easier to add newly mocked API requests, and rejects requests to algolia.io that are not mocked.

We've been using this change happily at Product Hunt for the last few months, sorry that it took so long to propose it upstream :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ index.set_settings({"customRanking" => ["desc(followers)"]})
 
 You can also configure the list of attributes you want to index by order of importance (first = most important):
 ```ruby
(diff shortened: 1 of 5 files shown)


lfittl:
@redox Sorry for the delay on this - I've now rebased and adjusted as we discussed previously.

redox:
Fixed it in #31 :+1:

lfittl:
@redox Awesome, appreciate it!

PR #11, opened 2015-03-02
Build with all Ruby 2.1.x versions

nicolasleger:


redox:
Let's see how it goes @nicolasleger :+1:

nicolasleger:
@redox We could also add build on ruby-head to test further incompatibility changes.
Should I add it?

PR #19, opened 2015-10-01
Update Algolia#search mocked response

mariusbutuc:
When testing `Algolia#search`, the response is different from what the docs suggest.

_(trailing whitespace trimmed as a side effect)_


redox:
LGTM :+1:

PR #24, opened 2015-12-11
Broken link to CDN

triplepointfive:
Same issue in documentation here https://www.algolia.com/doc/ruby#using-the-javascript-client

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ puts index.search('jim').to_json
   * It will offload unnecessary tasks from your servers
 
 ```html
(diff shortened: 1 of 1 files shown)


redox:
that's actually the expected URL. Without  specifying `http:` or `https:` it automatically uses the protocol of the current page. Your issue might be that you're testing a local file with `file://`, so it fails; isn't it?

triplepointfive:
Indeed, testing on local. It's a bit unexpectable since it's going to be tested on local first almost always, but OK, we can close it since it's not the typo.

PR #36, opened 2016-05-12
fixes browse when passing hash of params without block

robertjwhitney:


robertjwhitney:
Looks like the CI is broken.

robertjwhitney:
And pretty sure cover-rails is not up-to-date?

redox:
> Looks like the CI is broken.

Yeah, not exactly broken but since Travis doesn't allow PR to use the encrypted env variables, there aren't any Algolia credentials in the env when non-Algolia people submit PR. Not yet sure how better handling this but it's a pain... (cc @asurak)

Anyway, thanks for that; it makes a lot of sense to be able to use the new browse method without block. That being said, are you aware you're only get the first page of the browsing with such code?

Should we rather implement a `to_a` method on the IndexBrowser?

robertjwhitney:
Yeah I wouldn't expect it to return more than the page of results I asked for. This is helpful if you're trying to paginate beyond the first 1000 results. 

Also, if you are going to implement a `to_a` method, I wouldn't do it rather, but in addition to. All this really does is fix what looks like deprecated usage of `pageOrQueryParameters` in the `else` branch.

robertjwhitney:
Even that var name is kind of a smell, but I assume b/c supporting deprecated usage you can't do something like this?

`def browse(params = {}, &block)`

redox:
> Yeah I wouldn't expect it to return more than the page of results I asked for. This is helpful if you're trying to paginate beyond the first 1000 results.

Alright, let's do that then!

> I assume b/c supporting deprecated usage you can't do something like this?

Yes we would like to void breaking any API.

PR #41, opened 2016-05-18
Try to protect delete_by_query from wiping the whole index. Fix #40

### LINKED ISSUE #40: delete_by_query(nil) wipes entire index
`index.delete_by_query(nil)`

End result is 0 records in the relevant index.

redox:


rayrutjes:
This does help avoiding mass deletion to a limited extend, but it does introduce a BC issue. I doubt people are using such a way to clear the indices, but who knows. 
@dustincoates could you give us some context about your issue #40 ?

dustincoates:
@rayrutjes Sure thing. A customer brought up this problem where they are doing queries to figure out what to delete inside Algolia (because their indices mirror very closely their DBs). If a query returned `nil` they deleted their entire index so they had to do a lot of `nil` checks.

The mentioned that they were surprised by the behavior and found it unexpected (and, fwiw, I agree).

rayrutjes:
I totally agree with you and the customer, the behaviour is dangerous and tedious to protect. The best we can do from my POV is better document this method for now, and add a warning about the power of this feature. Or otherwise just introduce a minor BC, but not sure we wanna go down that road.

PR #96, opened 2016-10-20
feat(index): add search_facet

Jerskouille:
Adds the `search_facet` method to the index object.


redox:
Otherwise, 💯  @Jerskouille 👍

PR #109, opened 2016-12-02
Remember the last host used and re-use it if <60sec

redox:


redox:
@alexandremeunier @Jerskouille would love your review 👍

spec/client_spec.rb
```diff
@@ -997,6 +997,15 @@ def test_browse(expected, *args)
         @client.list_indexes # fallback on the second host after 5 sec (connection timeout)
         expect(start_time.to_i + 5).to be <= Time.now.to_i + 1
       end
+
+      it "should re-use the working the 2nd host after the first one failed" do
```
rayrutjes: Looks like you can remove 'the' before 2nd ;)

lib/algolia/client.rb
```diff
@@ -374,16 +377,27 @@ def delete(uri, type = :write)
 
     # This method returns a thread-local array of sessions
     def thread_local_hosts(read)
```
Jerskouille: I had a hard time following what would happen on the first call where current_index would be `nil`.
```
Thread.current[read ? :algolia_search_host_index : :algolia_host_index] ||= 0
```
would have made everything clearer here, and we could then remove the `.to_i` line 393.

Jerskouille:
LGTM 👍

lib/algolia/client.rb
redox: Arg yes indeed, using `nil.to_i` is something I've used to do.

PR #110, opened 2016-12-05
fix(hosts_keys): hosts_keys should be stored per app id

Jerskouille:
Right now, we can't have two clients with different app ids on the same thread.
This PR tries to fix this by replacing the `algolia{,_search}_hosts`  by `algolia{,_search}_hosts_${app_id}`.

Will impact #109 , which should use the same logic if we agree to go forward, be careful when merging.

Would love to have your input @redox and @alexandremeunier  !

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -41,6 +41,11 @@ module Algolia
       }
     end
 
+    def destroy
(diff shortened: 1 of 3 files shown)


redox:
Yep, that's great :+1: @Jerskouille ; could you then also update the `current_host` thread var to include that in  #109 and merge both?

PR #111, opened 2016-12-05
fix(client): client methods should be client-scoped

Jerskouille:
I'm not entirely sure if I got it all wrong, or if those weren't removed when it was supposed to be, can you confirm I'm not doing nonsense @redox ?


redox:
Arg wow, good catch @Jerskouille; not sure why this wasn't spot before.

Are you able to release a new patch version with those fixes?

Jerskouille:
Sure! I'll go forward then. 💯

PR #198, opened 2017-03-16
Add get_task_status method to Algolia::Index class.

srodman7689:
After looking through the source of this library I realized that there wasn't a way to just check the status of a task instead of waiting for a task to finish. So I decided to create a pull request in the hopes that you guys would merge it and add this functionality. I could really use it.

### DIFF
--- a/spec/client_spec.rb
+++ b/spec/client_spec.rb
@@ -691,6 +691,12 @@ describe 'Client' do
     [true, false].should include(task['pendingTask'])
   end
 
+  it 'Check attributes get_task_status' do
+    task = @index.add_object!({ :name => "John Doe", :email => "john@doe.org" }, "1")
+    status = @index.get_task_status(task["taskID"])
+    status.should be_a(String)
+  end
+
   it "Check add keys" do
     newIndexKey = @index.add_api_key(['search'])
     newIndexKey.should have_key('key')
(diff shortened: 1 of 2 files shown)


srodman7689:
I have created an issue in case you guys don't want to go this way with adding the functionality to get a tasks status here https://github.com/algolia/algoliasearch-client-ruby/issues/199

redox:
> After looking through the source of this library I realized that there wasn't a way to just check the status of a task instead of waiting for a task to finish. So I decided to create a pull request in the hopes that you guys would merge it and add this functionality. I could really use it.

Awesome @srodman7689 👍  Thank you so much for contributing!

spec/client_spec.rb
```diff
@@ -693,7 +693,7 @@ def wait_global_key_missing(key)
 
   it 'Check attributes get_task_status' do
     task = @index.add_object!({ :name => "John Doe", :email => "john@doe.org" }, "1")
-    status = @index.get_task_status(task["taskID"])
+    status = @index.get_task_status(task["objectID"])
```
redox: This was actually correct, we what to get the status of the `taskID` not the `objectID`.

redox:
Travis will fail because of the encrypted API keys we're using. I've run it locally 👍

PR #203, opened 2017-05-04
Remove Gemfile.lock

timcraft:
The Gemfile.lock file should not be checked into version control:

http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


raphi:
@timcraft can you update the Gemfile as well? Thx

timcraft:
@raphi Not sure what you mean, what needs updating in the Gemfile and how is that related to this pull request? The Gemfile.lock file doesn't need to be included in a gem.

raphi:
@timcraft sorry if it wasn't clear: I was referring to the part of adding `gemspec` in the Gemfile while removing dependencies duplicates.

Jerskouille:
Done in #257 . :)

PR #206, opened 2017-06-19
chore(browse): improve the default implementation

Jerskouille:
This should avoid the case where a user sets `distinct: 2` in his/her index settings and the API fail with `hitsPerPage * distinct should be <= 1000`.


Jerskouille:
This is useless, with `browse`, `distinct` is [already set to `false`](https://www.algolia.com/doc/api-client/php/advanced/#backup--export-an-index)

PR #208, opened 2017-06-22
Make delete_by_query async and add delete_by_query! to run synchronously

### LINKED ISSUE #204: Should there be a `delete_by_query!` method?
The `delete_object` and `delete_objects` methods both have a `!` version that will actually wait until the delete has taken place before returning. It appears that the `delete_by_query` method goes against this and waits by default. Is there any reason that this operation would need to wait?

obahareth:
This should fix #204.

### DIFF
--- a/lib/algolia/index.rb
+++ b/lib/algolia/index.rb
@@ -398,11 +398,7 @@ module Algolia
     #
     def delete_by_query(query, params = nil)
       raise ArgumentError.new('query cannot be nil, use the `clear` method to wipe the entire index') if query.nil? && params.nil?
-      params ||= {}
-      params.delete(:hitsPerPage)
-      params.delete('hitsPerPage')
-      params.delete(:attributesToRetrieve)
-      params.delete('attributesToRetrieve')
+      params = sanitized_delete_by_query_params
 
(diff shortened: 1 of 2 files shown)


obahareth:
The tests are failing because of missing a missing application ID and API key in the environment variables for Travis. I'm not sure how to fix that or I would.

```
/home/travis/build/algolia/algoliasearch-client-ruby/spec/spec_helper.rb:21:in `<top (required)>': missing ALGOLIA_APPLICATION_ID or ALGOLIA_API_KEY environment variables (RuntimeError)
```

obahareth:
I think perhaps there was a cleaner way to do this instead of the optional parameter? I wanted to avoid copy pasting the method because I wasn't sure how to deal with the loop, and this seemed the quickest not super-dirty way.

redox:
Thank you so much for your contribution @obahareth!

Unfortunately, this methods needs the `wait_task` in order to make the search results not return the objects that have just been deleted. I'll loop with @CBaptiste to double-check the C# code.

In the meantime, I've improved the code to avoid doing the `wait_task` if less than 1000 objects have been deleted. Let me know what you think of #209

PR #209, opened 2017-06-23
Improved delete_by_query to not run the wait_task if not appropriate

### LINKED ISSUE #204: Should there be a `delete_by_query!` method?

redox:
Fix #204

### DIFF
--- a/lib/algolia/index.rb
+++ b/lib/algolia/index.rb
@@ -411,9 +411,16 @@ module Algolia
       loop do
         res = search(query, params)
         break if res['hits'].empty?
+<<<<<<< HEAD
         last_task = delete_objects(res['hits'].map { |h| h['objectID'] })
         break if res['hits'].size < 1000
         wait_task(last_task['taskID'])
+=======
+        ids = res['hits'].map { |h| h['objectID'] }
+        res = delete_objects(ids)
+        break if ids.size != 1000 # no need to wait_task if we had less than 1000 objects matching; we won't get more after anyway.
+        wait_task res['taskID']
+>>>>>>> Improved delete_by_query to not run the wait_task if not appropriate
       end
       last_task
     end
--- a/lib/algolia/index.rb
+++ b/lib/algolia/index.rb
@@ -414,9 +414,10 @@ module Algolia
         result = browse(params)
 
         params[:cursor] = result['cursor']
-        break unless result.has_key?('hits')
 
         hits = result['hits']
+        break if hits.empty?
+
         ids += hits.map { |h| h['objectID'] }
       end
       last_task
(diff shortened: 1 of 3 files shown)


raphi:
👍

obahareth:
Hey, in my case I was deleting a large amount of objects (my records are paragraphs from large PDF files, and I don't really have an ID per paragraph stored anywhere) so I easily get more than 1000 objects. The delete by query literally took 15 mins+

Is there any other way we can speed it up for that scenario?

obahareth:
@redox Do you have any ideas or suggestions on how to handle a scenario like that?

lib/algolia/index.rb
```diff
@@ -409,7 +409,9 @@ def delete_by_query(query, params = nil)
       loop do
         res = search(query, params)
         break if res['hits'].empty?
-        res = delete_objects(res['hits'].map { |h| h['objectID'] })
+        ids = res['hits'].map { |h| h['objectID'] }
+        res = delete_objects(ids)
+        break if ids.size != 1000 # no need to wait_task if we had less than 1000 objects matching; we won't get more after anyway.
```
JanPetr: Shouldn't this condition be `continue if ids.size = 1000`?
If I get 1000 results it most probably means that I'll get another result with another search and there is no need for waiting for the delete task. Am I missing something?

JanPetr: So in the end the more results I get the longer I wait.

raphi: I think the problem here is that if you don't wait, the next `search()` calls might return the same object ids, thus flooding the API with duplicates delete calls.

raphi: One way to mitigate that would be to first scan the entire result set, and then calls to delete all of them.

JanPetr: Yes, didn't think of that. 
Changing implementation to you `browse` would break compatibility as the API key now needs only `search` and `delete records` ACLs. `browse` implementation would need `browse` ACL.

obahareth: @raphi By scanning the entire result set do you mean treating it as a paginated search and acquiring all the IDs of the objects to delete and then doing asynchronous deletes? I was about to try something like that.

raphi: @obahareth yes exactly. @redox what do you think?

obahareth: @raphi and @redox, I've played around with this in my own project and had very successful results, I tried deleting ~3000 objects with it and it happened within milliseconds instead of more than 5 minutes. I've opened PR #211 with the solution that worked for me based on @raphi's suggestion.

raphi:
🙌

raphi:
@obahareth I merge your PR here in order to trigger Travis's build

obahareth:
Thanks @raphi, I'm sorry there were still issues here and there.

obahareth:
I didn't get counted as a contributor 😟

Issue #33, opened 2016-04-22
browse documentation doesn't make any sense


vincentwoo:
Docs around 

```ruby
# Iterate with a filter over the index
index.browse({:query => "test", :numericFilters => 'i=42'}) do
    # Do something
end
```

need work - the block is passed something weird and is not explained

vincentwoo:
here's some terrifying shit:

```
42] pry(main)> index.browse.each { |hash| puts hash.class }; 2
Array
Array
Array
Array
Array
Array
Array
Array
Array
2
[43] pry(main)> index.browse.each { |arr, str| puts arr.class }; 2
String
String
String
String
String
String
String
String
String
2
[44] pry(main)> index.browse.each { |arr| puts arr.class }; 2
Array
Array
Array
Array
Array
Array
Array
Array
Array
2
[45] pry(main)> index.browse.each { |arr| puts arr[0] }; 2
hits
nbHits
page
nbPages
hitsPerPage
processingTimeMS
query
params
cursor
2
```

redox:
Yes, probably a lack of documentation here. Here is what the code is doing:

```ruby
            if block.arity == 2
              yield hit, @cursor
            else
              yield hit
            end
```

So you should do:

```ruby
index.browse({:query => "test", :numericFilters => 'i=42'}) do |hit|
    # Do something
end
```

or

```ruby
index.browse({:query => "test", :numericFilters => 'i=42'}) do |hit, cursor|
    # Do something
end
```

vincentwoo:
Come on, I'm not dumb. Your code is straight up broken and you probably need to rewrite your `browse` impl:

```
[5] pry(main)> index.browse({:query => "test"}) do |hit|
[5] pry(main)*   puts hit.class
[5] pry(main)* end
[algoliasearch-rails] Cannot GET to https://VXINIRAKQB-dsn.algolia.net/1/indexes/Question_development/browse?page=%7B%3Aquery%3D%3E%22test%22%7D&hitsPerPage=1000: {"message":"Invalid value for page parameter","status":400}

 (400)
[algoliasearch-rails] Cannot GET to https://VXINIRAKQB-dsn.algolia.net/1/indexes/Question_development/browse?page=%7B%3Aquery%3D%3E%22test%22%7D&hitsPerPage=1000: {"message":"Invalid value for page parameter","status":400}

 (400)
{
[...]

PR #211, opened 2017-07-24
Improve delete_by_query performance for queries that return more than 1000 hits

obahareth:
[As suggested](https://github.com/algolia/algoliasearch-client-ruby/pull/209#discussion_r128988549) by @raphi in #209, add a new version of `delete_by_query` that scans the entire index for all IDs to delete without awaiting delete results like in the old `delete_by_query` (which is now `delete_by_query!`) and then delete all matching objects asynchronously through one call to `delete_objects`.

This should solve #204 and provide a better-performing solution when dealing with a query that attempts to delete more than 1000 objects than my previous PR #208 and also #209.

### DIFF
--- a/lib/algolia/index.rb
+++ b/lib/algolia/index.rb
@@ -392,30 +392,31 @@ module Algolia
 
     #
     # Delete all objects matching a query
-    #
+    # This method retrieves all objects synchronously but deletes in batch
+    # asynchronously
     # @param query the query string
     # @param params the optional query parameters
     #
     def delete_by_query(query, params = nil)
       raise ArgumentError.new('query cannot be nil, use the `clear` method to wipe the entire index') if query.nil? && params.nil?
-      params ||= {}
-      params.delete(:hitsPerPage)
-      params.delete('hitsPerPage')
-      params.delete(:attributesToRetrieve)
-      params.delete('attributesToRetrieve')
+      params = sanitized_delete_by_query_params(params)
 
+      params[:query] = query
       params[:hitsPerPage] = 1000
       params[:distinct] = false
       params[:attributesToRetrieve] = ['objectID']
-      last_task = nil
-      loop do
-        res = search(query, params)
-        break if res['hits'].empty?
-        last_task = delete_objects(res['hits'].map { |h| h['objectID'] })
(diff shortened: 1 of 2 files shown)


raphi:
Hey, that looks good to me. However, I don't know how big the batch query payload can be. What if we have 10k, 100k, millions, ++ ids? We might need to batch them also, without waiting.

obahareth:
Cool, I would add that if I knew the limits for batch query payloads but I think this implementation should help people with scenarios like mine.

obahareth:
@raphi and @redox in regards to @JanPetr's [previous comment](https://github.com/algolia/algoliasearch-client-ruby/pull/209#discussion_r128990423) regarding breaking compatibility with ACLs due to requiring browse permissions, should I perhaps move my additions into new methods and leave the old ones untouched? That way we can retain the old version's compatibility while still providing a solution for users who would benefit from this enhancement.

obahareth:
Also looking at #33 it looks like I might be using `browse` wrong? There's no clear example for it in the docs. Should I use the version of `browse` that expects a block instead? could I get an example of how to iterate using it?

This is the code I tried locally before converting it to this library's conventions (with just a query, no filters) and it worked well:

```ruby
require 'algoliasearch'

application_id = "xxx"
api_key = "xxx"
Algolia.init(application_id: application_id, api_key: api_key)

index = Algolia::Index.new("people")

query = "john"

# Scan index
ids = []
cursor = ""

while cursor != nil
  result = index.browse({
    "query" => query,
    "params" => { "hitsPerPage" => 1000 },
    "cursor" => cursor })

  hits = result["hits"]
  cursor = result["cursor"]
  ids += hits.map { |h| h["objectID"] }

end

puts ids.length
index.delete_objects(ids)
```

raphi:
@redox as specified (here)[https://docs.travis-ci.com/user/encryption-keys/] it's not possible to test from forks.

I ran the tests manually to compare, fix a regression and now it's good to go 👍

raphi:
This PR was approved and merged in https://github.com/algolia/algoliasearch-client-ruby/pull/209 Thanks a lot @obahareth 👍

PR #213, opened 2017-07-28
feat(request_options): add request options to any method using API calls

Jerskouille:
Should be good now!

Please let's not write the associated tests for every method.

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -7,6 +7,7 @@ require 'openssl'
 require 'base64'
 
 module Algolia
+  WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY = 100
 
   # A class which encapsulates the HTTPS communication with the Algolia
   # API server. Uses the HTTPClient library for low-level HTTP communication.
@@ -162,7 +163,7 @@ module Algolia
     #
     def move_index!(src_index, dst_index, request_options = {})
       res = move_index(src_index, dst_index, request_options)
-      init_index(dst_index).wait_task(res['taskID'], 100, request_options)
+      init_index(dst_index).wait_task(res['taskID'], WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY, request_options)
       res
     end
 
@@ -188,7 +189,7 @@ module Algolia
     #
     def copy_index!(src_index, dst_index, scope = nil, request_options = {})
       res = copy_index(src_index, dst_index, scope, request_options)
-      init_index(dst_index).wait_task(res['taskID'], 100, request_options)
+      init_index(dst_index).wait_task(res['taskID'], WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY, request_options)
       res
     end
 
@@ -369,7 +370,7 @@ module Algolia
     def batch!(requests, request_options = {})
       res = batch(requests, request_options)
       res['taskID'].each do |index, taskID|
-        init_index(index).wait_task(taskID, 100, request_options)
+        init_index(index).wait_task(taskID, WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY, request_options)
       end
     end
 
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -261,7 +261,7 @@ module Algolia
     #  @param obj The list of parameters for this key.
     #         Defined by a Hash that can contain the following values:
     #          - acl: array of string
-    #          - indices: array of string
+    #          - indexes: array of string
     #          - validity: int
     #          - referers: array of string
     #          - description: string
@@ -311,7 +311,7 @@ module Algolia
     #  @param obj The list of parameters for this key.
     #         Defined by a Hash that can contain the following values:
     #          - acl: array of string
-    #          - indices: array of string
+    #          - indexes: array of string
     #          - validity: int
     #          - referers: array of string
     #          - description: string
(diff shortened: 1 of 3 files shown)


lib/algolia/client.rb
```diff
@@ -102,8 +102,9 @@ def with_rate_limits(end_user_ip, rate_limit_api_key, &block)
     # @param queries the array of hash representing the query and associated index name
     # @param index_name_key the name of the key used to fetch the index_name (:index_name by default)
     # @param strategy define the strategy applied on the sequential searches (none by default)
+    # @param request_options contains extra parameters to send with your query
     #
-    def multiple_queries(queries, index_name_key = :index_name, strategy = "none")
+    def multiple_queries(queries, index_name_key = :index_name, strategy = "none", request_options = {})
```
redox: Prototypes like this one really don't look ruby-ish :/

lib/algolia/client.rb
```diff
@@ -185,26 +192,31 @@ def delete_index!(name)
     #
     # @param offset Specify the first entry to retrieve (0-based, 0 is the most recent log entry).
     # @param length Specify the maximum number of entries to retrieve starting at offset. Maximum allowed value: 1000.
+    # @param request_options contains extra parameters to send with your query
     #
-    def get_logs(offset = 0, length = 10, type = "all")
+    def get_logs(offset = 0, length = 10, type = "all", request_options = {})
```
redox: Same :/

lib/algolia/client.rb
```diff
@@ -233,8 +245,9 @@ def get_api_key(key)
     #  @param maxQueriesPerIPPerHour the maximum number of API calls allowed from an IP address per hour (0 means unlimited)
     #  @param maxHitsPerQuery  the maximum number of hits this API key can retrieve in one call (0 means unlimited)
     #  @param indexes the optional list of targeted indexes
+    #  @param request_options contains extra parameters to send with your query
     #
-    def add_api_key(obj, validity = 0, maxQueriesPerIPPerHour = 0, maxHitsPerQuery = 0, indexes = nil)
+    def add_api_key(obj, validity = 0, maxQueriesPerIPPerHour = 0, maxHitsPerQuery = 0, indexes = nil, request_options = {})
```
redox: Same :/

lib/algolia/client.rb
```diff
@@ -413,16 +429,20 @@ def thread_local_hosts(read)
       end
     end
 
-    def perform_request(session, url, method, data)
+    def perform_request(session, url, method, data, request_options)
+      hs = {}
+      extra_headers = request_options[:headers] || request_options['headers'] || {}
+      @headers.each { |key, val| hs[key.to_s] = val }
+      extra_headers.each { |key, val| hs[key.to_s] = val }
```
redox: Why do you explicitly call `key.to_s`? Coud we write `hs = @headers.merge(request_options[:headers] || request_options['headers'] || {})` instead?

lib/algolia/index.rb
```diff
@@ -224,36 +236,39 @@ def get_object(objectID, attributesToRetrieve = nil)
...
 
     # Wait the publication of a task on the server.
     # All server task are asynchronous and you can check with this method that the task is published.
     #
     # @param taskID the id of the task returned by server
     # @param timeBeforeRetry the time in milliseconds before retry (default = 100ms)
+    # @param request_options contains extra parameters to send with your query
     #
-    def wait_task(taskID, timeBeforeRetry = 100)
+    def wait_task(taskID, timeBeforeRetry = 100, request_options = {})
```
redox: Same :/

lib/algolia/client.rb
Jerskouille: In order to make sure we're correctly handling `headers: { MyCustomHeader: 'xxx' }` instead of `headers: { 'MyCustomHeader' => 'xxx'}`.

lib/algolia/client.rb
Jerskouille: Agreed on all of those.
WDYT about:
```ruby
def multiple_queries(queries, options = nil, strategy = nil)
  if options.is_a?(Hash)
    index_name_key = options.delete(:index_name_key) || options.delete('index_name_key') || :index_name
    strategy = options.delete(:strategy) || options.delete('strategy') || 'none'
    request_options = options.delete(:request_options) || options.delete('request_options') || {}
  else
    index_name_key = options || :index_name
    strategy ||= 'none'
    request_options = {}
  end
  # ...
end
```

redox: Yes, that would be 💯

lib/algolia/client.rb
redox: But this will be automatically done by the underlying `httpclient` library, no?

Jerskouille: You're right, I had forgotten that ruby guaranteed the iteration on hashes to be in the order of insertion (JS doesn't).

lib/algolia/index.rb
Jerskouille: Here I would leave it like this. We also have `def add_object(obj, objectID = nil, request_options = {})` for instance.

If I do this, I should as well do this for each and every method.

Jerskouille:
Since this PR:
- delete_by_query was modified
- we've added query rules endpoints
- we've added delete_by

Jerskouille:
I've just rebased on master and fixed the conflicts.

I will review one last time myself and merge with my fingers crossed.
This PR has definitely been opened for too long.

PR #214, opened 2017-08-02
Fix a link to contacts.json

alsemyonov:


redox:
Good catch @alsemyonov; we've a README generator for all our libraries; I'll make sure it's updated there 👍

PR #222, opened 2017-08-17
Fine-tune the delete_by_query making it not wait_task by default.

### LINKED ISSUE #221: Ensure delete_by_query is mockable
From https://github.com/hannesfostie/big_articles/pull/1

redox:
Add also a delete_by_query! waiting for the last task.
Also set distinct to false.

Fix #221

### DIFF
--- a/lib/algolia/index.rb
+++ b/lib/algolia/index.rb
@@ -405,13 +405,28 @@ module Algolia
       params.delete('attributesToRetrieve')
 
       params[:hitsPerPage] = 1000
+      params[:distinct] = false
       params[:attributesToRetrieve] = ['objectID']
+      last_task = nil
       loop do
         res = search(query, params)
         break if res['hits'].empty?
-        res = delete_objects(res['hits'].map { |h| h['objectID'] })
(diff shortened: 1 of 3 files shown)


lib/algolia/index.rb
```diff
@@ -405,13 +405,28 @@ def delete_by_query(query, params = nil)
...
       params[:hitsPerPage] = 1000
+      params[:distinct] = false
       params[:attributesToRetrieve] = ['objectID']
+      last_task = nil
       loop do
         res = search(query, params)
         break if res['hits'].empty?
-        res = delete_objects(res['hits'].map { |h| h['objectID'] })
-        wait_task res['taskID']
+        last_task = delete_objects(res['hits'].map { |h| h['objectID'] })
```
raphi: Isn't it introducing duplicate requests? cf. convo https://github.com/algolia/algoliasearch-client-ruby/pull/211

redox: Arg, you're totally right.. didn't wake up while fixing this one.

PR #236, opened 2017-10-12
Allow to reuse the algolia webmocks

### LINKED ISSUE #100: Recommended Webmock integration is incorrect
Hey,

I think the recommended webmock setup in https://github.com/algolia/algoliasearch-client-ruby#mocking does not work in case of multiple test files because `require` will only load the code once. I think the stubs needs to be loaded in every test. Here is what I am doing instead:

```
RSpec.describe Event do

  specify do
    load $LOADED_FEATURES.find{|x| x.include?("algolia/version.rb") }.gsub("version.rb", "webmock.rb")

    # code which triggers algolia integration goes here....

    expect(WebMock).to have_requested(:put, "https://application_id.algolia.net/1/indexes/events/#{event.id}").
      with(body: hash_including({
        "objectID": event.id.to_s,
        "name": "dummy event",
        "category": "music",
      }))
  end

end
```

The key difference is that I do `load` instead of `require`. Unfortunately doing `load 'algolia/webmock'` did not just work so I had to use `$LOADED_FEATURES.find{|x| x.include?("algolia/version.rb") }.gsub("version.rb", "webmock.rb")` to find the file and load it.

I need to do it in every affected test.

### LINKED ISSUE #223: algolia/webmock does not correctly register request stubs
Hey folks,

I've mentioned this to @redox in a PR on one of our apps (or rather, a demo app) and have recreated the issue I've been experiencing in [this PR](https://github.com/hannesfostie/big_articles/pull/2)

The PR shows 2 commits, one being the initial setup for `webmock` and `webmock/algolia` where tests will fail. The second copies all request stubs from the file we required and adds them in the test file itself, making the tests pass.

TL;DR: just using `require 'webmock/algolia'` doesn't actually appear to register any stubs, and I'm not entirely sure why.

sobrinho:
This is needed because when you require algolia/webmock, the mocks will
work only in one spec at rspec, if you required webmock/rspec.

This happens because webmock/rspec will clear all stubs between examples
and we need to stub again for the next one.

The new approach is:

    require 'algolia/webmock'

    RSpec.configure do |config|
      config.before(:each) { Algolia::WebMock.mock! }
    end

Fixes #223 
Fixes #100

### DIFF
--- a/lib/algolia/webmock.rb
+++ b/lib/algolia/webmock.rb
@@ -5,42 +5,48 @@ rescue LoadError
   exit 1
 end
 
-# list indexes
-WebMock.stub_request(:get, /.*\.algolia(net\.com|\.net)\/1\/indexes/).to_return(:body => '{ "items": [] }')
-# query index
-WebMock.stub_request(:get, /.*\.algolia(net\.com|\.net)\/1\/indexes\/[^\/]+/).to_return(:body => '{ "hits": [ { "objectID": 42 } ], "page": 1, "hitsPerPage": 1, "nbHits": 1, "nbPages": 1 }')
-# delete index
-WebMock.stub_request(:delete, /.*\.algolia(net\.com|\.net)\/1\/indexes\/[^\/]+/).to_return(:body => '{ "taskID": 42 }')
-# clear index
-WebMock.stub_request(:post, /.*\.algolia(net\.com|\.net)\/1\/indexes\/[^\/]+\/clear/).to_return(:body => '{ "taskID": 42 }')
-# add object
-WebMock.stub_request(:post, /.*\.algolia(net\.com|\.net)\/1\/indexes\/[^\/]+/).to_return(:body => '{ "taskID": 42 }')
-# save object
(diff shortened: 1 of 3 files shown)


raphi:
LGTM, locally tested, all tests passed 👍

redox:
Do you guys think such a "breaking change" is acceptable without increasing the major version of the gem? @raphi @sobrinho ? Because it's test, I feel like it's fine.

sobrinho:
To keep the backward compatibility we can call `Algolia::WebMock.mock!` at the end of the file.

In that way, this will be an incremental change and we can remove that call in the next major version to be incompatible one.

This assuming that we do not have changes to justify a major version but I'm okay with both approaches.

oehlschl:
@raphi @sobrinho are there any additional changes required here? We'd benefit from seeing this merged as well.

sobrinho:
I made the change incremental now, the old behavior will be kept by then not needing a major release.

We may remove that call from the next major version but isn't required at all.

sobrinho:
Travis is red because of missing ALGOLIA_APPLICATION_ID or ALGOLIA_API_KEY environment variables.

Can you run locally again @raphi and merge?

raphi:
Tests passed. @sobrinho thanks for your contribution and patience^^ 👍

PR #240, opened 2017-11-10
Make HttpClient transparent_gzip_decompression configurable

pimpin:
Transparent gzip decompression in not compatible with httplog gem.

Fix https://github.com/algolia/algoliasearch-rails/issues/273

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -24,6 +24,7 @@ module Algolia
 
       @ssl             = data[:ssl].nil? ? true : data[:ssl]
       @ssl_version     = data[:ssl_version].nil? ? nil : data[:ssl_version]
+      @transparent_gzip_decompression = data[:transparent_gzip_decompression].nil? ? true : data[:transparent_gzip_decompression]
       @application_id  = data[:application_id]
       @api_key         = data[:api_key]
       @hosts           = data[:hosts] || (["#{@application_id}.algolia.net"] + 1.upto(3).map { |i| "#{@application_id}-#{i}.algolianet.com" }.shuffle)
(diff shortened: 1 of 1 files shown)


pimpin:
You might be right, I did not check in the HttpClient source yet.
But it is only for dev purpose.
Could we add it to configuration set ? Do you have other suggestion to make it compatible with httplog ?

Jerskouille:
I believe that if you were to expose a client configuration option as for `@ssl = data[:ssl].nil? ? true : data[:ssl]`, this should be considerable cc @redox ?

redox:
> I believe that if you were to expose a client configuration option as for @ssl = data[:ssl].nil? ? true : data[:ssl], this should be considerable cc @redox ?

Yes, this could be an option. We could have a `:gzip` option.

pimpin:
Thanks for suggestion @Jerskouille . @redox Could you implement it please ?
I'll be happy to test the branch as soon as possible and give you any feedback nedded.

Jerskouille:
Actually @pimpin , since this is your PR, I believe you should be able to modify your code without too much hassle. :)

pimpin:
@Jerskouille I first did not see the link with `@ssl = data[:ssl].nil? ? true : data[:ssl]` . But I think I get it.

I did not try this yet. I will tonight.
@redox could you confirm if this is seems ok for you please ?

If my issue is fixed, I might also try to add a spec and squash this.

pimpin:
I've tried it and it works for us.
I've rebased and skip the spec part :innocent: 
Is this ok for you ? @redox

Jerskouille:
Can you maybe call it only `@gzip`? Otherwise, I believe we're good to go, and I'd be happy to merge it.

pimpin:
I'm not really aware of the reason why this param is called `transprent_gzip_compression` in HttpClient ; But I do believe in sharpness over shortness importance when dealing with naming.
But it's your choice and it's ok for me.

Jerskouille:
To answer your comment, I agree with you.
However in this case, setting this to false effectively disables `gzip` entirely for the client, hence making a good case, at least for me, for a short yet clear option.

PR #243, opened 2017-11-15
feat(copyIndexScope): adding scopes to `copy_index`

raphi:


spec/client_spec.rb
```diff
@@ -293,6 +293,31 @@ def is_include(array, attr, value)
...
+      { :objectID => 'city', :type => 'synonym', :synonyms => ['San Francisco', 'SF'] },
+      { :objectID => 'street', :type => 'altCorrection1', :word => 'street', :corrections => ['st'] }
+    ]
+    @index.search('')['nbHits'].should eq(1)
+    @index.search_synonyms('')['nbHits'].should eq(2)
+
+    res = Algolia.copy_index!(safe_index_name("àlgol?a"), safe_index_name("àlgol?à"), ["synonyms"])
+
+    @index.delete_index!
+
```
Jerskouille: Shouldn't you check here that `@index.search('')['nbHits'].should eq(0)` ?

raphi: :D nice catch. Indeed that's what I did at first but this is triggering an error from the API due to a weird behavior that won't be fixed.

Jerskouille:
Do you have right to merge and deploy @raphi ? If yes, I believe we're good, feel free to merge and release. :)

raphi:
@Jerskouille yes I do, will do thanks :)

PR #245, opened 2017-11-21
Remove rubygems requires

timcraft:
They aren't necessary:

http://www.rubyinside.com/why-using-require-rubygems-is-wrong-1478.html

https://github.com/rails/rails/commit/d8c194968469f3afe39412818505ac94f78e386d#diff-64230db87218c6b7c92fb9b6c4436c46

Fewer lines of code to maintain 😊


raphi:
https://github.com/algolia/algoliasearch-client-ruby/pull/256

PR #257, opened 2017-12-06
Remove Gemfile.lock

raphi:


algoliasearch.gemspec
```diff
@@ -53,15 +53,18 @@ Gem::Specification.new do |s|
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
       s.add_runtime_dependency(%q<httpclient>, ["~> 2.8.3"])
       s.add_runtime_dependency(%q<json>, [">= 1.5.1"])
+      s.add_runtime_dependency(%q<hashdiff>, "0.3.5")
```
benhutton: @raphi why did this need to get added here? And does it need such a strict version specification? And does it need to use an older version of the gem? (A newer version has been out for 2 months! https://github.com/liufengyun/hashdiff/blob/master/changelog.md#v037-2017-10-08)

raphi: @benhutton this is for Ruby 1.8 compatibility

benhutton: @raphi are you using it anywhere? Github search isn't returning anything.

It just feels a little... heavy to lock a gem version this tightly, especially to support a version of ruby that was retired 4.5 years ago (https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/). Hashdiff not supporting ruby 1.8 is a perfectly normal thing within the ruby ecosystem. Locking gems so tightly to preserve 1.8.7 support is _not_ a normal thing. One hard version lock like this can have a large cascading effect on a Gemfile, holding back other upgrades.

PR #262, opened 2017-12-14
Fix `hashdiff` wrong dependency

raphi:
Attempt to fix hardcoded dependency due to `hashdiff` wrong dependency declaration (https://github.com/algolia/algoliasearch-client-ruby/pull/257/files/36bcd0b1c4d05776dcbdb362c15a609c81f41cde)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,9 +5,10 @@ git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
 # Load algoliasearch.gemspec dependencies
 gemspec
 
-# if RUBY_VERSION < '1.9.3'
-#   gem 'hashdiff', '< 0.3.6' # Hashdiff 0.3.6 no longer supports Ruby 1.8
-# end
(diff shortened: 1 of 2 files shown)


raphi:
cc @benhutton

Gemfile
```diff
@@ -5,6 +5,11 @@ git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
 # Load algoliasearch.gemspec dependencies
 gemspec
 
+# See https://github.com/algolia/algoliasearch-client-ruby/pull/257/files/36bcd0b1c4d05776dcbdb362c15a609c81f41cde
+if RUBY_VERSION < '1.9.3'
```
Jerskouille: Let's maybe avoid a check on strings directly.
`Gem::Version.new` works perfectly for that:

https://github.com/algolia/algoliasearch-rails/blob/2db3c5886421b7667b1c864466f4e8195ad1a343/Gemfile#L30-L35

PR #263, opened 2017-12-15
Add gemspec metadata

raphi:
Simple refactor mainly adding metadata for [RubyGems](https://rubygems.org/gems/algoliasearch)


raphi:
@Jerskouille `~> 2.8.3` < `2.9` while `~> 2.8` < `3.0`
So in that case, we allow minor backward compatible upgrades but with the bug fix `2.8.3`

PR #273, opened 2018-03-14
Add support of HTTP_PROXY environment variable

vinc:
This PR adds support of `HTTP_PROXY` environment variable to the gem like in the JavaScript client: https://github.com/algolia/algoliasearch-client-javascript#proxy-support

If the environment variable is not set, the code behave as usual.

See https://github.com/nahi/httpclient/blob/v2.8.3/sample/thread.rb#L7

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -439,7 +439,7 @@ module Algolia
     def thread_local_hosts(read)
       thread_hosts_key = read ? "algolia_search_hosts_#{application_id}" : "algolia_hosts_#{application_id}"
       Thread.current[thread_hosts_key] ||= (read ? search_hosts : hosts).each_with_index.map do |host, i|
-        client = HTTPClient.new
+        client = HTTPClient.new(ENV["HTTP_PROXY"])
         client.ssl_config.ssl_version = @ssl_version if @ssl && @ssl_version
         client.transparent_gzip_decompression = @gzip
         client.ssl_config.add_trust_ca File.expand_path(File.join(File.dirname(__FILE__), '..', '..', 'resources', 'ca-bundle.crt'))
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ index.set_settings searchableAttributes: %w(
 )
 ```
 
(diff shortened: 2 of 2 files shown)


vinc:
Any tips on how to test this feature?

julienbourdeau:
Good question :D I was about to ask you ^^ I'll think about it, it would be great to add tests.

vinc:
I'll see if I can find some inspiration from this: https://github.com/nahi/httpclient/blob/v3.2.8/test/helper.rb#L69

lib/algolia/client.rb
```diff
@@ -439,7 +439,7 @@ def delete(uri, type = :write, request_options = {})
     def thread_local_hosts(read)
       thread_hosts_key = read ? "algolia_search_hosts_#{application_id}" : "algolia_hosts_#{application_id}"
       Thread.current[thread_hosts_key] ||= (read ? search_hosts : hosts).each_with_index.map do |host, i|
-        client = HTTPClient.new
+        client = HTTPClient.new(ENV["HTTP_PROXY"])
```
redox: This shouldn't be needed because the `httpclient` library already read this variables:
 - documented in https://github.com/nahi/httpclient/blob/master/lib/httpclient.rb#L46-L47
 - loaded from https://github.com/nahi/httpclient/blob/master/lib/httpclient.rb#L1065-L1077
 - called from https://github.com/nahi/httpclient/blob/master/lib/httpclient.rb#L449

Any chance you could tell us what happens in this `load_environment` method in your case @vinc ? Maybe you have a `REQUEST_METHOD` variable forcing the code to look for `cgi_http_proxy` ?

vinc: Indeed yesterday our fork seemed to fix the issue we had on our production server, but we switched back to the original gem after reading your comment and everything is still working.

Therefor the issue we were having seems to have been totally unrelated!

So we can close this PR, and I apologize for your time spent on this :3

raphi: No problem @vinc thanks for contributing!

PR #282, opened 2018-04-09
UserAgent: Add ruby version

julienbourdeau:
We'd like to add the language version to each client. Today everything is overridden if you pass a user-agent value to the constructor of the Client class. 
I'd rather leave the client and language version and let users add something specific at the end.

Any comment is welcome.

Pull request on Rails is coming.

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -21,7 +21,7 @@ module Algolia
     DEFAULT_SEND_TIMEOUT    = 30
     DEFAULT_BATCH_TIMEOUT   = 120
     DEFAULT_SEARCH_TIMEOUT  = 5
-    DEFAULT_USER_AGENT      = ["Ruby (#{RUBY_VERSION})", "Algolia for Ruby (#{::Algolia::VERSION})"]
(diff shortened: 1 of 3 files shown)


lib/algolia/client.rb
```diff
@@ -490,6 +490,12 @@ def perform_request(session, url, method, data, request_options)
       return JSON.parse(response.content)
     end
 
+    def format_user_agent(custom)
```
raphi: It works well, but I think we could simply do a constant for UA to avoid useless string allocations.
Something like:
```
[...]
'User-Agent' => USER_AGENT + data[:user_agent].to_s
[...]
```

raphi:
cc @julienbourdeau should be good but just to confirm: the syntax of the UA is changing, even for custom UA, is that what you want?

julienbourdeau:
I have checked in the dashboard and the client version must be passed first, I updated the code.

Please see https://gist.github.com/julienbourdeau/7c11df8647e85f5fc7eac9d2c017e52c

If the tests pass, the PR is ready to merge 🎉

raphi:
@julienbourdeau can you rebase it on master to get the MCM tests fixed?

PR #283, opened 2018-04-11
fix: Prevent saving a rule with an empty objectID

aseure:
First time for years I'm writing Ruby. Absolutely any comment is more than welcome.


aseure:
Follow up of https://github.com/algolia/algoliasearch-client-go/issues/397.

PR #286, opened 2018-04-30
feat: Deprecate API keys methods on Index in favor of Client ones

aseure:


raphi:
@aseure how about we start to warn dev but output warning on console/logs as well?

raphi:
Sure but (almost) nobody reads the raw client code. Most big gems do warn that way, and it's personally the only way I can see it and then do the upgrade.

julienbourdeau:
About adding proper warnings, this is definitely something we want to start doing (in all languages).
We'll open a PR to add warnings on all deprecated method 👍

PR #289, opened 2018-05-07
Utils: move to CHANGELOG.md and add Rake task for release

raphi:


raphi:
Thanks! Pretty stupid/basic so far but does the job!

PR #290, opened 2018-05-15
Fix CHANGELOG link in gemspec

Spone:


raphi:
Nice catch! Thanks for contributing

Issue #269, opened 2018-02-01
Client review


julienbourdeau:
List of things to change (if possible)

- [x] `index.search` rename params to searchParameters
- [x] `search_for_facet_value` rename parameters
- [x] Fix browse (and doc snippets)
- [x] `delete_objects` pass objectIds list
- [x] `wait_task` timeBeforeRetry arg ?
- [x] `get_settings` remove options?
- [x]  `Algolia.init` refactor ENV['xxx'] ?

raphi:
@julienbourdeau what are the valid Algolia ENV vars? `ALGOLIA_API_ID`, `ALGOLIA_APPLICATION_ID`, any others?

julienbourdeau:
* `ALGOLIA_APP_ID` or `ALGOLIA_APPLICATION_ID` for the app id (use short version if possible)
* `ALGOLIA_API_KEY` for the almost-admin key
* `ALGOLIA_SEARCH_API_KEY` for the search only key

Do you need places or MCM credentials as well?

raphi:
Alright thanks I will add `ALGOLIA_APP_ID` in the ruby client then. No I don't need the places one for this client; MCM maybe when we will add it :trollface:

PR #292, opened 2018-05-15
API review

raphi:
https://github.com/algolia/algoliasearch-client-ruby/issues/269 PR that went under the radar cc @julienbourdeau

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -9,8 +9,10 @@ require 'base64'
 module Algolia
   WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY = 100
 
+  #
   # A class which encapsulates the HTTPS communication with the Algolia
   # API server. Uses the HTTPClient library for low-level HTTP communication.
+  #
   class Client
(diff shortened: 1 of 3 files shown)


lib/algolia/client.rb
```diff
@@ -697,32 +719,34 @@ def Algolia.get_user_key(key, request_options = {})
...
-  #          - maxHitsPerQuery: integer
+  #          - max_hits_per_query: integer
   #          - queryParameters: string
-  #          - maxQueriesPerIPPerHour: integer
+  #          - max_queries_per_IP_per_hour: integer
   #  @param request_options contains extra parameters to send with your query - Default = {}
   #
-  def Algolia.add_api_key(obj, request_options = {}, maxQueriesPerIPPerHour = 0, maxHitsPerQuery = 0, indexes = nil)
-    Algolia.client.add_api_key(obj, request_options, maxQueriesPerIPPerHour, maxHitsPerQuery, indexes)
+  def Algolia.add_api_key(object, request_options = {}, max_queries_per_IP_per_hour = 0, max_hits_per_query = 0, indexes = nil)
```
julienbourdeau: Isn't renaming optional parameter a break of backward compat'? I might be mixing up with Python.

raphi: Nope, you can even do `add_api_key(plop = 'hello')` it will work; the order is important though

PR #293, opened 2018-05-16
Update common files

julienbourdeau:
:gift:


raphi:
Actually @julienbourdeau it doesn't look good when you open an issue/PR:

screen shot 2018-05-25 at 2 39 51 pm

PR #295, opened 2018-05-25
Fix(travis): always test on the latest patches

raphi:


raphi:
For sure, it's not necessary! Those two versions are memorable / well known in the community, are on end of support and the last ones they will ever release (they might do security fixes but on these patch versions directly)

PR #296, opened 2018-05-28
Fix(syntax): this isn't PHP

raphi:
Remove useless `;`


raphi:
@julienbourdeau can you fix the MCM tests please? All PRs are failing now

PR #299, opened 2018-06-01
Tests: make MCM tests more stable

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## What was changed

I introduce a new helper functions in test called `auto_retry`. Some operations are asynchronous but the engine doesn't return a taskID. This function execute a block and retry multiple times before failing.

For MCM, if a userID wasn't assigned before we try to retrieve or delete, it would fail. With `auto_retry` we wait a few seconds and try agian before failing.

Another thing, when you search for a userID, you might not find the one you want because there is one with a close name ranked higher. I fixed that by using a unique userID for every build.

## Why it was changed

New MCM tests were making the build very unstable.

### DIFF
--- a/spec/client_mcm_spec.rb
+++ b/spec/client_mcm_spec.rb
@@ -46,11 +46,13 @@ describe 'Multi Cluster Management', :mcm => true do
 
   it 'should assign a user to a cluster' do
     @client.assign_user_id(@user_id, @cluster_name)
-    sleep(2)
   end
 
   it 'should get the created user id' do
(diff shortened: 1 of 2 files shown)


julienbourdeau:
No it's like adding a new Key. It's asynchronous but without a taskID. That's why I went for this loop. I retry a few times and wait to see if the task could be complete. After a while the operation as probably failed for real.

PR #303, opened 2018-06-14
Feat/ab testing

julienbourdeau:
See PR on PHP client for the details: https://github.com/algolia/algoliasearch-client-php/pull/408

Note that this should be merged after #302

### DIFF
--- a/lib/algolia/analytics.rb
+++ b/lib/algolia/analytics.rb
@@ -1,0 +1,75 @@
+module Algolia
+
+  class Analytics
+    attr_reader :ssl, :ssl_version, :headers
+    API_URL='https://analytics.algolia.com'
+
+    def initialize(client, params)
+      @client   = client
+      @headers  = params[:headers]
+    end
+
+    def get_ab_tests(params = {})
+      params = {
+        :offset => 0,
+        :limit => 10,
+      }.merge(params)
+
+      perform_request(:GET, Protocol.ab_tests_uri, params)
+    end
+
+    def get_ab_test(ab_test_id)
+      raise ArgumentError.new('ab_test_id cannot be empty') if ab_test_id.nil? || ab_test_id == ''
+
+      perform_request(:GET, Protocol.ab_tests_uri(ab_test_id))
+    end
+
+    def add_ab_test(ab_test)
+      perform_request(:POST, Protocol.ab_tests_uri, {}, ab_test.to_json)
+    end
+
+    def stop_ab_test(ab_test_id)
+      raise ArgumentError.new('ab_test_id cannot be empty') if ab_test_id.nil? || ab_test_id == ''
+
+      perform_request(:POST, Protocol.ab_tests_stop_uri(ab_test_id))
+    end
+
+    def delete_ab_test(ab_test_id)
+      raise ArgumentError.new('ab_test_id cannot be empty') if ab_test_id.nil? || ab_test_id == ''
+
+      perform_request(:DELETE, Protocol.ab_tests_uri(ab_test_id))
+    end
+
+    def wait_task(index_name, taskID, time_before_retry = WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY, request_options = {})
+      @client.wait_task(index_name, taskID, time_before_retry, request_options)
+    end
+
+    private
+
+    def perform_request(method, url, params = {}, data = {})
+      http = HTTPClient.new
+
+      url = API_URL + url
+
+      encoded_params = Hash[params.map { |k, v| [k.to_s, v.is_a?(Array) ? v.to_json : v] }]
+      url << "?" + Protocol.to_query(encoded_params)
+
+      response = case method
+                 when :GET
+                   http.get(url, { :header => @headers })
+                 when :POST
+                   http.post(url, { :body => data, :header => @headers })
+                 when :DELETE
+                   http.delete(url, { :header => @headers })
+                 end
+
+      if response.code / 100 != 2
+        raise AlgoliaProtocolError.new(response.code, "Cannot #{method} to #{url}: #{response.content}")
+      end
+
+      JSON.parse(response.content)
+    end
+
+  end
+
+end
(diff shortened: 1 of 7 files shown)


lib/algolia/analytics.rb
```diff
@@ -0,0 +1,77 @@
+module Algolia
+
+  class Analytics
+    attr_reader :ssl, :ssl_version, :headers
+
+    def initialize(client, params)
+      @client = client
+
+      @headers      = params[:headers]
+      @ssl          = params[:ssl].nil? ? true : params[:ssl]
```
raphi: It is easier to read: `@ssl = params[:ssl] || true`

lib/algolia/analytics.rb
```diff
@@ -0,0 +1,77 @@
...
+      @ssl_version  = params[:ssl_version].nil? ? nil : params[:ssl_version]
```
raphi: I'm not sure to understand what's the intent here? In any case, it is equal to `params[:ssl_version]`. So this should be enough: `@ssl_version = params[:ssl_version]`

lib/algolia/analytics.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    def get_ab_tests(params = {})
+      params = {
+          :offset => 0,
```
raphi: Indentation should be 2 spaces in Ruby

lib/algolia/analytics.rb
```diff
@@ -0,0 +1,77 @@
...
+          :limit => 10,
+      }.merge(params)
+
+      perform_request(:GET, Protocol.ab_tests_uri, params)
+    end
+
+    def get_ab_test(ab_test_id)
+      raise ArgumentError.new('ab_test_id cannot be empty') if ab_test_id.nil? || ab_test_id == ''
```
raphi: Maybe `ab_test_id.empty?` instead of `''`?

lib/algolia/analytics.rb
```diff
@@ -0,0 +1,77 @@
...
+      perform_request(:DELETE, Protocol.ab_tests_uri(ab_test_id))
+    end
+
+    def wait_task(index_name, taskID, time_before_retry = WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY, request_options = {})
+      @client.wait_task(index_name, taskID, time_before_retry, request_options)
+    end
+
+    private
+
+    def perform_request(method, url, params = {}, data = {})
```
raphi: Is there a reason why we duplicate this method and don't use the client's one? We could also leverage the thread pool by using the client's.

lib/algolia/analytics.rb
```diff
@@ -0,0 +1,77 @@
...
+      client = HTTPClient.new
+
+      url.prepend('https://analytics.algolia.com')
```
raphi: Ahah remember `unshift`^^ ?

raphi: Use const

lib/algolia/client.rb
```diff
@@ -168,7 +179,7 @@ def move_index(src_index, dst_index, request_options = {})
     #
     def move_index!(src_index, dst_index, request_options = {})
       res = move_index(src_index, dst_index, request_options)
-      init_index(dst_index).wait_task(res['taskID'], WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY, request_options)
+      wait_task(dst_index, res['taskID'], WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY, request_options)
```
raphi: Woops I think you `checkout` on the wrong branch :/

lib/algolia/analytics.rb
julienbourdeau: If you use `init_analytics` yes, but if you build it yourself, you might not pass it. Also, at the the time I wrote this, I didn't pass the client to analytics. Maybe I could just read these value from the client 🤔

lib/algolia/analytics.rb
julienbourdeau: 👍 definitely!
I copy-pasted the init function of the client 😅

julienbourdeau:
About wait_task, I rebase the branch on top of #302 for Travis. I'll merge #302 first, then rebase this one onto master 👌

lib/algolia/analytics.rb
julienbourdeau: Do you mean moving `'https://analytics.algolia.com'` to a class const?

lib/algolia/analytics.rb
raphi: Yep, it's not necessary a reference 😅

lib/algolia/analytics.rb
raphi: Yeah, if you don;t pass it `@ssl_version = params[:ssl_version]` then `@ssl_version` will be `nil`, exactly like `params[:ssl_version].nil? ? nil`

lib/algolia/analytics.rb
julienbourdeau: I dont think we can easily use the `client.perform_request` without `client.request`. And `client.request` is coupled to `@hosts`, I don't see how to override the host when calling `request()`  without modifying the client instance.

Although, I might be wrong, I'm not sure if Ruby passed a reference to the same client instance or a copy 🤔

Note that for analytics, we don't need the retry strategy nor timeouts.

raphi: Yeah indeed it requires some adjustments on the client but it's cleaner than copying the logic IMO. It's always by "reference" kind of. You need to manually `.dup` object otherwise.

lib/algolia/analytics.rb
julienbourdeau: I checked, it seems like it's only for array. In this case URL is a string.

lib/algolia/analytics.rb
julienbourdeau: It doesn't work because ab_test_id is normally a `fixnum`

lib/algolia/analytics.rb
julienbourdeau: It seems like a fairly big refactor to do. I'd rather do it later if necessary. Currently, analytics api doesn't need timouts nor retry strategy.

PR #311, opened 2018-07-06
Prepend custom user agent

### LINKED ISSUE #310: Custom user agent is displayed in wrong position
- Algolia Client Version: 1.23.0
- Language Version: all

### Description
Fix for issue on Dashboard where the wrong User agent was being displayed (Rails should be listed first instead of Ruby)

`algolia_for_ruby_(1.23.2);_ruby_(2.3.1);_algolia_for_rails_(1.20.4);_algolia_for_ruby_(1.23.2)`

amcdaid106:
This should fix this issue https://github.com/algolia/AlgoliaWeb/issues/3515#issuecomment-402992627 by making sure the custom user agent is displayed first

| Q                 | A
| Bug fix?          | yes
| New feature?      | no  
| BC breaks?        | no     
| Related Issue     | Fix #310
| Need Doc update   | no

## What was changed
Prepends `data[:user_agent]` instead of appending it to `DEFAULT_USER_AGENT`

## Why it was changed
This behavior corrects expected order from Dashboard and possibly other systems

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -42,7 +42,7 @@ module Algolia
         Protocol::HEADER_API_KEY => api_key,
         Protocol::HEADER_APP_ID  => application_id,
         'Content-Type'           => 'application/json; charset=utf-8',
-        'User-Agent'             => DEFAULT_USER_AGENT.prepend(data[:user_agent]).compact.join('; ')
+        'User-Agent'             => DEFAULT_USER_AGENT.unshift(data[:user_agent]).compact.join('; ')
       }
     end
 
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -42,7 +42,7 @@ module Algolia
         Protocol::HEADER_API_KEY => api_key,
         Protocol::HEADER_APP_ID  => application_id,
         'Content-Type'           => 'application/json; charset=utf-8',
-        'User-Agent'             => DEFAULT_USER_AGENT.push(data[:user_agent]).compact.join('; ')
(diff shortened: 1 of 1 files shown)


raphi:
IMO it make sense and it's a clean fix. The only thing I see, is that if we have some users using custom headers, it will be displayed before the Ruby one. If that's ok with the rest of the systems then why not.

For example, it would become `my_custom_user_agent customer Blabla; Rails ; Ruby`, are we ok with that @julienbourdeau ?

The only change @amcdaid106 is to use `unshift` instead of `prepend`, because `prepend` was added for Ruby 2, while `unshift` is compatible with other versions of Ruby.

amcdaid106:
Ok @raphi changed it to `unshift` 🙌

julienbourdeau:
It's displayed in this order in all other clients. I'll have a deeper look into this. First, I need to check why we have underscores `_` instead of spaces and see if we need to change it here, in Rails, or in all clients.

I have written this doc when I updated UserAgent recently: https://gist.github.com/julienbourdeau/7c11df8647e85f5fc7eac9d2c017e52c

julienbourdeau:
In the end, it's impossible to keep the correct order in the client-side. We may want to add the version of an integration using this client, which should be prepended because it's more important, by maybe you also want to add another plugin you use, which should be appened.

We tried with PHP (where we have the most integrations and smaller helper libs) and the conclusion is that the normalization must be done on our side, not in the clients. This will also help for all the clients not up-to-date sending old UserAgent.

I'm going to work on this with Raymond on Q1 👌

PR #335, opened 2019-01-18
Add restore_api_key method

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

Algolia API introduced a new endpoint to restore a deleted API key 🎉 

```ruby
Algolia.restore_api_key(the_deleted_key, request_options)
```

### DIFF
--- a/lib/algolia/client.rb
+++ b/lib/algolia/client.rb
@@ -456,6 +456,13 @@ module Algolia
     end
 
     #
+    # Restore a deleted api key
+    #
+    def restore_api_key(key, request_options = {})
+      post(Protocol.restore_key_uri(key), :write, request_options)
+    end
+
+    #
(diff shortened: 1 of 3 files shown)


lib/algolia/client.rb
```diff
@@ -455,6 +455,13 @@ def delete_api_key(key, request_options = {})
       delete(Protocol.key_uri(key), :write, request_options)
     end
 
+    #
+    # Restore a deleted api key
+    #
+    def restore_api_key(key, request_options = {})
+      post(Protocol.key_uri(key), :write, request_options)
```
nunomaduro: Should not this be `restore_key_uri` instead `key_uri`?

PR #337, opened 2019-01-22
Fix flaky AB Test tests

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

AB Test tests started failing randomly, I think this is due to some change on the Analytics API.

Instead of only waiting on the taskID, I also pull the data until it's confirmed to be applied (like keys or MCM operations)


julienbourdeau:
This was fixed on the API side 🙆‍♂️

PR #359, opened 2019-08-19
added(index): unit testing relevance helpers

### LINKED ISSUE #356: Implement two new methods for unit testing of relevance
- [Specs and CTS tests](https://github.com/algolia/algoliasearch-client-specs-internal/pull/41)
- [Go client base implementation](https://github.com/algolia/algoliasearch-client-go/pull/524)

aseure:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Close #356 
| Need Doc update   | no

@nunomaduro @chloelbn When you'll come back, I would really appreciate some guidance from you on this PR. 99% of the job is done (implementation + tests are the same as Python/Go implementations) but I'd need some help regarding the following:

The Ruby `1.8.7` Travis build which does not pass. It seems that `rvm`, which is used to install/select the Ruby version in Travis environment is not able to install the version. I've tried their command on a fresh Linux install in Docker and I can reproduce (i.e. `rvm use 1.8.7-p371 --install --binary --fuzz` does not work). Should we drop the support for this version then or tell Travis about it, I did not found any related issue? For now, I've made this builid part of the `allowed failures` category in case we want to move forward.


lib/algolia/index.rb
```diff
@@ -264,6 +264,56 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
       client.post(Protocol.objects_uri, { :requests => requests }.to_json, :read, request_options)['results']
     end
 
+    #
```
nunomaduro: Those docs explain how the method is implemented, and not how should people work with the method. Check the other examples on the file.

lib/algolia/index.rb
```diff
@@ -264,6 +264,56 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+    # If no object has been found within the first result set, the function
+    # will perform a new search operation on the next page of results, if any,
+    # until a matching object is found or the end of results, whichever
+    # happens first.
+    #
+    # To prevent the iteration through pages of results, `paginate`
+    # parameter can be set to false. This will stop the function at the end of
+    # the first page of search results even if no object does match.
+    #
+    def find_object(filter_func, request_options = {})
```
nunomaduro: Is `filter_func` idiomatic in ruby?

nunomaduro: We may have to adapt this method, as the `search` as search_parameters and request_options. Any idea?

lib/algolia/index.rb
```diff
@@ -264,6 +264,56 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+                'object' => hit,
+                'position' => i,
+                'page' => page,
+            }
+          end
+        end
+
+        has_next_page = page + 1 < res['nbPages']
+        if not paginate or not has_next_page
+          raise AlgoliaError.new('object not found')
```
nunomaduro: Can't we create a specific exception for this?

lib/algolia/index_helpers.rb
```diff
@@ -0,0 +1,12 @@
+module Algolia
+
+  def self.get_object_position(res, object_id)
```
nunomaduro: I would prefer to see this method as static method on the `search_index`.

lib/algolia/index_helpers.rb
```diff
@@ -0,0 +1,12 @@
...
+    res['hits'].each_with_index do |hit, i|
+      if hit['objectID'] == object_id
+        return i
+      end
+    end
+    -1
+  end
+
+end
```
nunomaduro: Missing EOL

lib/algolia/error.rb
```diff
@@ -20,4 +20,18 @@ def initialize(code, message)
     end
   end
 
-end
\ No newline at end of file
+  # An exception class raised when the given object was not found.
```
nunomaduro: This should be another exception like on the remaining validity pr.

nunomaduro:
@julienbourdeau Can you help me to review this? Not the logic, but the ruby idiomatics.

lib/algolia/index.rb
```diff
@@ -264,6 +264,72 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+    # The function takes a block to filter the results from search query
+    # Usage example:
+    #  index.find_object({'query' => '', 'paginate' => true}) {|obj| obj.key?('company') and obj['company'] == 'Apple'}
+    #
+    # @param request_options contains extra parameters to send with your query
+    #
+    def find_object(request_options = {})
+      query = ''
+      paginate = true
+      page = 0
```
raphi: The 2 first initializations are useless since you override it just below. If both `request_options[:query] || request_options['query']` are nil, `query` will be nil and not `''`. Is that what you expected?

Not sure how `search(nil)` is handle behind, it might be fine?

Worst case, you could do `query = request_options[:query] || request_options['query'] || ''`

lib/algolia/index.rb
```diff
@@ -264,6 +264,72 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+            return {
+                'object' => hit,
+                'position' => i,
+                'page' => page,
+            }
+          end
+        end
+
+        has_next_page = page + 1 < res['nbPages']
+        if not paginate or not has_next_page
```
raphi: I would highly recommend to not use `not` and `or` unless really necessary. It is not an equivalent to the bang `!` or `||`.

See the exact behaviour in this cool example: https://stackoverflow.com/questions/27429711/difference-between-not-and-in-ruby?answertab=votes#tab-top

lib/algolia/index.rb
```diff
@@ -264,6 +264,72 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+      paginate = request_options[:paginate] || request_options['paginate']
+      request_options.delete(:paginate)
+      request_options.delete('paginate')
+
+      while true
+        request_options['page'] = page
+        res = search(query, request_options)
+
+        res['hits'].each_with_index do |hit, i|
+          if yield(hit)
```
raphi: Looks good 👍 

The only issue I see is if the user didn't pass a block as expected, there will be a `LocalJumpError (no block given (yield))` error.

Maybe that's expected? If not, you can use `if block_given?`

lib/algolia/index.rb
```diff
@@ -264,6 +264,72 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+    # @return [Integer] position of the object, or -1 if it's not in the array
+    #
+    def self.get_object_position(res, object_id)
+      res['hits'].each_with_index do |hit, i|
+        if hit['objectID'] == object_id
+          return i
+        end
+      end
+      -1
+    end
```
raphi: It works!

I would rename `res` to something more readable like `hits`, `objects` or `records`.

Also, a more Rubyiesque way could be:
```ruby
def self.get_object_position(res, object_id)
  res['hits'].find_index { |hit| hit['objectID'] == object_id } || -1
end
```
Double check the Ruby version though :)

spec/client_spec.rb
```diff
@@ -446,6 +446,66 @@ def wait_global_key_missing(key)
...
+    }.to raise_exception(
+      Algolia::AlgoliaObjectNotFoundError,
+      'Object not found'
+    )
+
+    obj = index.find_object({'query' => '', 'paginate' => true, 'hitsPerPage' => 5}, &condition)
+    obj['position'].should eq(0)
+    obj['page'].should eq(2)
+  end
+
```
raphi: Could be good to add a test without a block passed

lib/algolia/index.rb
```diff
@@ -264,6 +264,73 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+      request_options.delete(:query)
+      request_options.delete('query')
+
+      if request_options.has_key? :paginate
+        paginate = request_options[:paginate]
+      end
+
+      if request_options.has_key? 'paginate'
+        paginate = request_options['paginate']
+      end
```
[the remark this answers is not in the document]
raphi: Duh. There is probably a shorter way but it's readable as is.

lib/algolia/index.rb
```diff
@@ -264,6 +264,73 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+    #
+    # Options can be passed in request_options body:
+    #  - query (string): pass a query
+    #  - paginate (bool): choose if you want to iterate through all the
+    # documents (true) or only the first page (false). Default is true.
+    # The function takes a block to filter the results from search query
+    # Usage example:
+    #  index.find_object({'query' => '', 'paginate' => true}) {|obj| obj.key?('company') and obj['company'] == 'Apple'}
+    #
+    # @param request_options contains extra parameters to send with your query
```
nunomaduro: No return type?

lib/algolia/index.rb
```diff
@@ -264,6 +264,73 @@ def get_objects(objectIDs, attributes_to_retrieve = nil, request_options = {})
...
+    #
+    def find_object(request_options = {})
+      paginate = true
+      page = 0
+
+      query = request_options[:query] || request_options['query'] || ''
+      request_options.delete(:query)
+      request_options.delete('query')
+
+      if request_options.has_key? :paginate
```
nunomaduro: Can we have the same syntax here?

```ruby
paginate = request_options[:paginate] || request_options['paginate'] || true
request_options.delete(:paginate)
request_options.delete('paginate')
```

PR #361, opened 2019-09-04
feat: remaining validity

### LINKED ISSUE #357: New method: GetSecuredApiKeyRemainingValidity
## Summary

When a secured API keys expire, search breaks. The current workaround is more complex
than it needs to be. 

The options for developers using Algolia require, at every place where a search
operation is performed, catching an operation error, identify it as a
"ValidUntil" error, then regenerate the API key, and perform the search
operation again.

The idea is to define a method which tells if a `SecuredAPIKey`
is expired or not without making a single call to Algolia's API.

## References

- Specs: https://github.com/algolia/algoliasearch-client-specs-internal/pull/25

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #357
| Need Doc update   | yes

### DIFF
--- a/lib/algolia/error.rb
+++ b/lib/algolia/error.rb
@@ -20,4 +20,18 @@ module Algolia
     end
   end
 
+  # An exception class raised when the validUntil parameter is not found
+  # The error code and message will be parsed out of the HTTP response,
+  # which is also included in the response attribute.
+  class ValidUntilNotFoundError < AlgoliaError
+    attr_accessor :code
+    attr_accessor :message
+
+    def initialize(code, message)
+      self.code = code
+      self.message = message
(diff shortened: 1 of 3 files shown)


lib/algolia/error.rb
```diff
@@ -20,4 +20,18 @@ def initialize(code, message)
     end
   end
 
-end
\ No newline at end of file
+  # An exception class raised when the validUntil parameter is not found
+  # The error code and message will be parsed out of the HTTP response,
+  # which is also included in the response attribute.
+  class ValidUntilNotFoundError < AlgoliaError
+    attr_accessor :code
```
nunomaduro: There is no status code in this exception.

PR #362, opened 2019-09-09
fix(mcm-tests): update tests to make them pass

chloelbn:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

Fixes failing MCM tests

### DIFF
--- a/spec/client_mcm_spec.rb
+++ b/spec/client_mcm_spec.rb
@@ -33,9 +33,14 @@ describe 'Multi Cluster Management', :mcm => true do
     entry.count.should eq(4)
   end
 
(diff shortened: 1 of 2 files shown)


spec/client_mcm_spec.rb
```diff
@@ -34,8 +34,14 @@ def safe_user_id(name)
   end
 
   it 'should list user ids' do
+    new_user_id1 = safe_user_id('ruby-client-2')
```
nunomaduro: Maybe `ruby-client-1` ?

PR #364, opened 2019-09-19
fix(exists-method): fix method syntax with an alias

### LINKED ISSUE #363: Fix: update index.exists method syntax
`index.exists` method returns a boolean and thus should have the following syntax => `index.exists?`. Create an alias to fix this.

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #363
| Need Doc update   | no

### DIFF
--- a/lib/algolia/index.rb
+++ b/lib/algolia/index.rb
@@ -1295,6 +1295,11 @@ module Algolia
       return true
     end
 
+    #
+    # Aliases the exists method
+    #
+    alias :exists? :exists
+
     # Deprecated
     alias_method :get_user_key, :get_api_key
     alias_method :list_user_keys, :list_api_keys
--- a/spec/client_spec.rb
+++ b/spec/client_spec.rb
@@ -177,12 +177,12 @@ describe 'Client' do
 
   it "should tell if index exists" do
     @index.add_object!({ :name => "John Doe", :email => "john@doe.org" }, "1")
-    expect(@index.exists).to be true
+    expect(@index.exists?).to be true
   end
 
   it "should tell if index does not exist" do
     index = Algolia::Index.new('nonexistent_index')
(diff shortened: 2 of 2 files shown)


lib/algolia/index.rb
```diff
@@ -1295,6 +1295,11 @@ def exists
       return true
     end
 
+    #
+    # Aliases the exists method
+    #
+    alias exists? exists
```
raphi: It's valid, I just have a preference for `alias :exists? :exists` as it shows a clear reference to a method. But this is nitpicking.

spec/client_spec.rb
```diff
@@ -177,12 +177,12 @@ def wait_global_key_missing(key)
 
   it "should tell if index exists" do
     @index.add_object!({ :name => "John Doe", :email => "john@doe.org" }, "1")
-    expect(@index.exists).to be true
+    expect(@index.exists?).to be true
```
raphi: Do we want to test both since both are valid?

lib/algolia/index.rb
chloelbn: I thought that you had to use the `:` only with `alias_method`. Would you favor `alias_method`?

spec/client_spec.rb
chloelbn: I thought that only `exists?` would need to be tested since this is the recommended usage. what do you think?

lib/algolia/index.rb
raphi: There are slight differences between the two of them but in this case both works. No the `:` works as well for `alias`

spec/client_spec.rb
raphi: Up to your team's good practices, how do you do in other clients?

PR #378, opened 2020-02-05
fix(search_user_ids): rename parameter clusterName

### LINKED ISSUE #368: Ensure `searchUserIDs()` accepts `cluster` and not `clusterName` parameter
The Algolia documentation was wrongly documenting that the `searchUserIDs()`
method is accepting an optional `clusterName` parameter. Actually, the
parameter can only be passed as `cluster`.

chloelbn:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #368
| Need Doc update   | yes

## Describe your change

The method `search_user_id` expected an optional parameter `clusterName`. Actually, the expected parameter is `cluster`

### DIFF
--- a/algoliasearch.gemspec
+++ b/algoliasearch.gemspec
@@ -64,8 +64,13 @@ Gem::Specification.new do |s|
     s.specification_version = 4
 
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
+      if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new('1.9.3')
+        s.add_runtime_dependency     'json',       '>= 1.5.1', '< 2.3'
+      else
+        s.add_runtime_dependency     'json',       '>= 1.5.1'
+      end
+
       s.add_runtime_dependency     'httpclient', '~> 2.8', '>= 2.8.3'
-      s.add_runtime_dependency     'json',       '>= 1.5.1', '< 2.3'
       s.add_development_dependency 'travis',     '~> 0'
       s.add_development_dependency 'rake',       '~> 0'
       s.add_development_dependency 'rdoc',       '~> 0'
--- a/algoliasearch.gemspec
+++ b/algoliasearch.gemspec
@@ -65,16 +65,16 @@ Gem::Specification.new do |s|
 
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
       s.add_runtime_dependency     'httpclient', '~> 2.8', '>= 2.8.3'
-      s.add_runtime_dependency     'json',       '>= 1.5.1'
+      s.add_runtime_dependency     'json',       '>= 1.5.1', '< 2.3'
       s.add_development_dependency 'travis',     '~> 0'
       s.add_development_dependency 'rake',       '~> 0'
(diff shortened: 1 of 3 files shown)


lib/algolia/client.rb
```diff
@@ -552,7 +552,7 @@ def remove_user_id(user_id, request_options = {})
 
     def search_user_id(query, cluster_name = nil, page = nil, hits_per_page = nil, request_options = {})
       body = { :query => query }
-      body[:clusterName] = cluster_name unless cluster_name.nil?
```
nunomaduro: What is happening today? This argument don't work at all?

chloelbn: Indeed, the engine throws an error `{"message":"invalid key 'clusterName'. Expected: query, cluster, appID, apiKey, page, hitsPerPage (near 1:49)","status":400}`

nunomaduro:
Hey @chloelbn, since this is a bug, I am gonna lets you decide if you wanna ship this bug fix.

algoliasearch.gemspec
```diff
@@ -65,16 +65,16 @@ Gem::Specification.new do |s|
 
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
       s.add_runtime_dependency     'httpclient', '~> 2.8', '>= 2.8.3'
-      s.add_runtime_dependency     'json',       '>= 1.5.1'
+      s.add_runtime_dependency     'json',       '>= 1.5.1', '< 2.3'
```
nunomaduro, line 68: Can this change effect current customers?

algoliasearch.gemspec
```diff
@@ -65,16 +65,16 @@ Gem::Specification.new do |s|
...
       s.add_development_dependency 'travis',     '~> 0'
       s.add_development_dependency 'rake',       '~> 0'
       s.add_development_dependency 'rdoc',       '~> 0'
     else
       s.add_dependency             'httpclient', '~> 2.8', '>= 2.8.3'
-      s.add_dependency             'json',       '>= 1.5.1'
+      s.add_dependency             'json',       '>= 1.5.1', '< 2.3'
```
nunomaduro, line 74: Can this change effect current customers?

algoliasearch.gemspec
```diff
@@ -65,16 +65,16 @@ Gem::Specification.new do |s|
...
     end
   else
     s.add_dependency               'httpclient', '~> 2.8', '>= 2.8.3'
-    s.add_dependency               'json',       '>= 1.5.1'
+    s.add_dependency               'json',       '>= 1.5.1', '< 2.3'
```
nunomaduro, line 78: Can this change effect current customers?

algoliasearch.gemspec
chloelbn: indeed, this has been fixed

algoliasearch.gemspec
chloelbn: in this case and the next one, it targets older versions of Ruby which would have the same dependency problem so we can block the version

algoliasearch.gemspec
chloelbn: see previous comment

PR #399, opened 2020-08-28
fix(account-client): retrieve all objects from source index

### LINKED ISSUE #398: Copy Index from One App to Another Only Sends Over 1000 Records
- Algolia Client Version: #.#.#
- Language Version:

### Description
Using algoliasearch-rails (1.23.2)

In the Account Client copy_index method to copy an index from one app to another app, the copy_index method iterates over the returned array from the index's browse function. It then works in batches of 1000 to save the objects. However, when I run the copy_index function on our source index that contains 64,000 records, I only ever receive 1000 objects in my destination index. It seems like the iterating over the batches is broken.

### Steps To Reproduce
```
prod_client = Algolia::Client.new({
  application_id: id,
  api_key: api key
})
staging_client = Algolia::Client.new({
  application_id: id,
  api_key: api key
})

prod_reseller_item = prod_client.init_index('prod_index')
staging_client.delete_index!('staging_index')
staging_reseller_item = staging_client.init_index('staging_index')

Algolia::AccountClient.copy_index(prod_reseller_item, staging_reseller_item)

The copy_index then returns only 1000 objects.
```

chloelbn:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #398
| Need Doc update   | no

## Describe your change

There was a bug preventing the fetching of all the objects from the source index. Instead, it retrieved only the first 1000 as it's the default number of hits per page. This commit fixes it

### DIFF
--- a/spec/account_client_spec.rb
+++ b/spec/account_client_spec.rb
@@ -47,9 +47,7 @@ describe 'Account client' do
 
   it 'should perform a cross app copy index and assert that destination must not exist' do
 
-    @index_1.save_objects! ([
-        {:objectID => 'one'},
-    ])
+    @index_1.save_objects!(1.upto(1500).map { |i| { :objectID => i, :i => i } })
 
     @index_1.batch_rules! ([
         {
@@ -67,9 +65,9 @@ describe 'Account client' do
 
     Algolia::AccountClient.copy_index!(@index_1, @index_3)
 
-    res = @index_3.search('');
-    res['nbHits'].should eq(1)
-    res['hits'][0]['objectID'].should eq('one')
+    res = @index_3.search('')
+    res['nbHits'].should eq(1500)
+    res['hits'][0]['objectID'].should eq('999')
 
     res = @index_3.search_rules('')['hits']
     res.size.should eq(1)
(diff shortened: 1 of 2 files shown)


redox (approved):
LGTM

spec/account_client_spec.rb
```diff
@@ -67,9 +65,9 @@
 
     Algolia::AccountClient.copy_index!(@index_1, @index_3)
 
-    res = @index_3.search('');
-    res['nbHits'].should eq(1)
-    res['hits'][0]['objectID'].should eq('one')
+    res = @index_3.search('')
+    res['nbHits'].should eq(1500)
+    res['hits'][0]['objectID'].should eq('999')
```
redox, line 70: I think this test is confusing, since the index is not ordered; we don't know why '999' is first. Let's remove this line.

chloelbn: What about adding a comment above the line then? At the end of the day I'd like to test that we retrieve the same objects and this is one way of doing it. 
Or I can save one last object with a clearer ID, change the `hitsPerPage` and retrieve this object instead

redox: if so, let's rather compare the whole list of 1500 objects then :) ensuring all numbers are retrieved.

redox (approved):
LGTM!

aseure (approved):
✅ LGTM

jbiffis:
Hi @redox @aseure @chloelbn  just wondering when this will get merged.  We're held up on some work on our end waiting for this.

PR #400, opened 2020-09-03
Chore/automate release process

aseure:
**Summary:**

This PR originally intended to automate the RubyGems build and publish steps
into CircleCI. However, because this release step is dependent on the tests to
pass, I had to fix here and there, which have been adressed in different
commits.

The release process has been updated in our Confluence space, the credentials
have been safely stored (see our Confluence page for more details) and I've
udpated our CircleCI configuration with the new RubyGems token as a new secured
environment variable (which remains not shared with forked PR builds).

**Testing plan:**

All tests and rubocop formatting checks are still expected to pass like before.
I've also added some unit tests for a new added function which are also
expected to pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,8 +110,8 @@ workflows:
       - test_ruby:
           matrix:
             parameters:
-              version: ['2.2', '2.3', '2.4', '2.5', '2.6', 'latest']
+              version: ['2.2', '2.3', '2.4', '2.5', '2.6', '2.7']
       - test_jruby:
           matrix:
             parameters:
-              version: ['9', '9.1', '9.2', 'latest']
+              version: ['9.1', '9.2']
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,13 +29,6 @@ aliases:
        mkdir /tmp/test-results
        bundle exec rake test:all
 
-  - &store_test_results
-    path: /tmp/test-results
-
-  - &store_artifacts
-    path: /tmp/test-results
-    destination: test-results
-
 references:
   default_docker_ruby_executor: &default_docker_ruby_executor
     image: circleci/ruby:<< parameters.version >>
@@ -81,8 +74,6 @@ jobs:
       - run: *install_bundler
       - save_cache: *save_cache
       - run: *run_tests
-      - store_test_results: *store_test_results
-      - store_artifacts: *store_artifacts
 
   test_jruby:
     description: Build, unit and integration tests for JRuby
@@ -98,8 +89,6 @@ jobs:
       - run: *install_bundler
       - save_cache: *save_cache
       - run: *run_tests
-      - store_test_results: *store_test_results
-      - store_artifacts: *store_artifacts
 
   release:
     description: Release a new version of the Ruby client
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,11 +21,7 @@ aliases:
 
   - &run_lint
     name: Run linting tool
-    command: |
-      if [ << parameters.version >> -eq '2.6' || 'latest' ]
-      then
-        bundle exec rake rubocop
-      fi
+    command: bundle exec rake rubocop
 
   - &run_tests
      name: Run unit and integration tests
@@ -58,6 +54,19 @@ references:
 version: 2.1
 
 jobs:
+  format:
+    description: Run our linter checks against the entire code base
+    parameters:
+      version:
+        type: string
+    docker:
+      - *default_docker_ruby_executor
+    steps:
+      - checkout
+      - run: *check_bundler
+      - run: *install_bundler
+      - run: *run_lint
+
   test_ruby:
     description: Build, unit and integration tests for Ruby
     parameters:
@@ -71,7 +80,6 @@ jobs:
       - restore_cache: *restore_cache
       - run: *install_bundler
       - save_cache: *save_cache
-      - run: *run_lint
       - run: *run_tests
       - store_test_results: *store_test_results
       - store_artifacts: *store_artifacts
@@ -89,7 +97,6 @@ jobs:
       - restore_cache: *restore_cache
       - run: *install_bundler
       - save_cache: *save_cache
-      - run: *run_lint
       - run: *run_tests
       - store_test_results: *store_test_results
       - store_artifacts: *store_artifacts
@@ -98,6 +105,8 @@ workflows:
   version: 2
   ci:
     jobs:
+      - format:
+          version: '2.7'
       - test_ruby:
           matrix:
             parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,26 @@ jobs:
       - store_test_results: *store_test_results
       - store_artifacts: *store_artifacts
 
+  release:
+    description: Release a new version of the Ruby client
+    parameters:
+      version:
+        type: string
+    docker:
+      - *default_docker_ruby_executor
+    steps:
+      - checkout
+      - run: *check_bundler
+      - run: *install_bundler
+      - run:
+          name: Build new gem
+          command: gem build algolia.gemspec
+      - run:
+          name: Publish new gem
+          command: |
+            if [[ -z "$RUBYGEMS_API_KEY" ]]; then echo '$RUBYGEMS_API_KEY is not set'; exit 1; fi
+            gem push --key $RUBYGEMS_API_KEY $(ls algolia-*.gem)
+
 workflows:
   version: 2
   ci:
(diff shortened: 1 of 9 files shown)


test/algolia/unit/helpers_test.rb
```diff
@@ -21,4 +22,48 @@ def test_deserialize_settings
       assert_equal new_settings, deserialized_settings
     end
   end
+
+  describe 'test hash_includes_subset' do
+    def test_empty_hashes
+      h      = {}
+      subset = {}
+      assert_equal true, hash_includes_subset?(h, subset)
```
chloelbn, line 30:
```suggestion
      assert hash_includes_subset?(h, subset)
```

test/algolia/unit/helpers_test.rb
```diff
@@ -21,4 +22,48 @@ def test_deserialize_settings
...
+    end
+
+    def test_with_empty_subset
+      h      = { a: 100, b: 200 }
+      subset = {}
+      assert_equal true, hash_includes_subset?(h, subset)
```
chloelbn, line 36:
```suggestion
      assert hash_includes_subset?(h, subset)
```

test/algolia/unit/helpers_test.rb
```diff
@@ -21,4 +22,48 @@ def test_deserialize_settings
...
+    end
+
+    def test_subset_included
+      h      = { a: 100, b: 200 }
+      subset = { a: 100 }
+      assert_equal true, hash_includes_subset?(h, subset)
```
chloelbn, line 42:
```suggestion
      assert hash_includes_subset?(h, subset)
```

.circleci/config.yml
```diff
@@ -89,20 +97,52 @@ jobs:
...
   version: 2
   ci:
     jobs:
+      - format:
+          version: '2.7.1'
       - test_ruby:
           matrix:
             parameters:
-              version: ['2.2', '2.3', '2.4', '2.5', '2.6', 'latest']
+              version: ['2.2.10', '2.3.8', '2.4.10', '2.5.8', '2.6.6', '2.7.1']
```
DevinCodes, line 133: Why do we have the version fixed to e.g. `2.5.8` instead of `2.5`? If I remember correctly, using `2.5` will always take the latest version of `2.5.x`, so if there's (for example) a security flaw that is fixed and `2.5.9` is released, we'll test against that version instead of `2.5.8`.

test/algolia/unit/helpers_test.rb
```diff
@@ -21,4 +22,48 @@ def test_deserialize_settings
...
+    end
+
+    def test_subset_not_included
+      h      = { a: 100, b: 200 }
+      subset = { c: 300 }
+      assert_equal false, hash_includes_subset?(h, subset)
```
chloelbn, line 48:
```suggestion
      refute hash_includes_subset?(h, subset)
```

test/algolia/unit/helpers_test.rb
```diff
@@ -21,4 +22,48 @@ def test_deserialize_settings
...
+    end
+
+    def test_subset_included_but_wrong_value
+      h      = { a: 100, b: 200 }
+      subset = { a: 200 }
+      assert_equal false, hash_includes_subset?(h, subset)
```
chloelbn, line 54:
```suggestion
      refute hash_includes_subset?(h, subset)
```

test/algolia/unit/helpers_test.rb
```diff
@@ -21,4 +22,48 @@ def test_deserialize_settings
...
+    end
+
+    def test_subset_included_with_multiple_values
+      h      = { a: 100, b: 200, c: 300 }
+      subset = { a: 100, b: 200 }
+      assert_equal true, hash_includes_subset?(h, subset)
```
chloelbn, line 60:
```suggestion
      assert hash_includes_subset?(h, subset)
```

test/algolia/unit/helpers_test.rb
```diff
@@ -21,4 +22,48 @@ def test_deserialize_settings
...
+    end
+
+    def test_subset_not_included_because_too_many_values
+      h      = { a: 100, b: 200 }
+      subset = { a: 100, b: 200, c: 300 }
+      assert_equal false, hash_includes_subset?(h, subset)
```
chloelbn, line 66:
```suggestion
      refute hash_includes_subset?(h, subset)
```

.circleci/config.yml
```diff
@@ -89,20 +97,52 @@ jobs:
...
             parameters:
-              version: ['9', '9.1', '9.2', 'latest']
+              version: ['9.1.17', '9.2.13']
+      - release:
+          version: '2.7.1'
+          requires:
+            - format
+            - test_ruby
+            - test_jruby
+          filters:
```
DevinCodes, line 144: Given that CircleCI treats these filters as `OR` filters, would having only the `tags` filter work for us?

chloelbn (changes requested):

.circleci/config.yml
```diff
@@ -89,20 +97,52 @@ jobs:
...
+      - run: *check_bundler
+      - run: *install_bundler
+      - run:
+          name: Build new gem
+          command: gem build algolia.gemspec
+      - run:
+          name: Publish new gem
+          command: |
+            if [[ -z "$RUBYGEMS_API_KEY" ]]; then echo '$RUBYGEMS_API_KEY is not set'; exit 1; fi
+            gem push --key $RUBYGEMS_API_KEY $(ls algolia-*.gem)
```
chloelbn, line 122: Could you add a step to delete the built gem as well please?

chloelbn (changes requested):

lib/algolia/responses/update_api_key_response.rb
```diff
@@ -23,8 +24,8 @@ def wait(opts = {})
       until @done
         begin
           api_key = @client.get_api_key(@raw_response[:key], opts)
-          @done   = @request_options <= api_key
-        rescue StandardError => e
+          @done   = hash_includes_subset?(api_key, @request_options)
+        rescue AlgoliaError => e
           if e.code != 404
             raise e
           end
```
DevinCodes, line 31:
```suggestion
          raise e unless e.code == 404
```

test/algolia/integration/recommendation_client_test.rb
```diff
@@ -17,7 +17,13 @@ def test_recommendation_client
         personalizationImpact: 0
       }
 
-      client.set_personalization_strategy(personalization_strategy)
+      begin
+        client.set_personalization_strategy(personalization_strategy)
+      rescue Algolia::AlgoliaHttpError => e
+        if e.code != 429
+          raise e
+        end
```
DevinCodes, line 25:
```suggestion
        raise e unless e.code == 429
```

DevinCodes (changes requested):
Thank you for this, just a few remarks but it looks good to me! 💯

.circleci/config.yml
DevinCodes: I don't think that's necessary, given that we don't store the cache during this job

aseure: Yes, I had the same reasoning as @DevinCodes because the cache is not saved and the Docker instance will get trashed anyway, I did not bother making the configuration longer.

.circleci/config.yml
aseure: Just fixed it in a separate commit (to keep the history if someone is looking at the blame one day).

.circleci/config.yml
aseure: Since I wasn't sure the latest patch version was used, I've preferred to set them explicitly. This indeed has the drawback you mention of not updating to the latest possible version.

I know maintainers have different views regarding the explicit vs. implicit declarations so I'll let you @DevinCodes and @chloelbn tell me what you prefer here, since you'll be the maintainers, and I'll either keep or remove the commit accordingly.

aseure:
Thanks a lot for your reviews @DevinCodes and @chloelbn! Just waiting for your latest answers and we're good to go 🚀

.circleci/config.yml
DevinCodes: My vote goes to implicit declarations here.

What about you @chloelbn ?

chloelbn: Now that the problem with the older versions is fixed I'd rather go back to implicit as well

chloelbn (approved):

.circleci/config.yml
aseure: Just changed it. Thanks for your reviews 👍 

I've kept `2.7` instead of `latest` to prevent us from having to look on Docker Hub to understand which version is used. Same for JRuby.

DevinCodes (approved):

PR #401, opened 2020-09-03
feat(steep): add iterators

chloelbn:
Adds typing of the iterators folder + Helpers class

### DIFF
--- a/sig/helpers.rbs
+++ b/sig/helpers.rbs
@@ -1,7 +1,7 @@
 module Helpers
 
   def to_json: (String|Hash[Symbol, String] body) -> String
-  def symbolize_hash: (Hash[Symbol|String, untyped] hash) -> Hash[Symbol, String]
+  def symbolize_hash: (Hash[Symbol|String, untyped] hash) -> Hash[Symbol, untyped]
   def handle_params: (Hash[Symbol, String] params) -> String
   def to_query_string: (Hash[Symbol, String] params) -> String
   def json_to_hash: (String json, bool symbolize_keys) -> Hash[Symbol, String]
--- a/sig/helpers.rbs
+++ b/sig/helpers.rbs
@@ -1,35 +1,12 @@
 module Helpers
-  # Convert an Hash to json
-  #
+
   def to_json: (String|Hash[Symbol, String] body) -> String
-
-  # Converts each key of a hash to symbols
-  #
-  def symbolize_hash: (Hash[Symbol|String, String] hash) -> Hash[Symbol, String]
-
-  # Convert params to a full query string
-  #
+  def symbolize_hash: (Hash[Symbol|String, untyped] hash) -> Hash[Symbol, String]
   def handle_params: (Hash[Symbol, String] params) -> String
-
-  # Create a query string from params
-  #
   def to_query_string: (Hash[Symbol, String] params) -> String
-
-  # Convert a json object to an hash
-  #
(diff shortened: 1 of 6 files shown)


sig/helpers.rbs
```diff
@@ -0,0 +1,35 @@
+module Helpers
+  # Convert an Hash to json
```
DevinCodes, line 2: I think we can remove all the comments in all the `.rbs` files (except the TODO items). It's just extra stuff to maintain, wdyt?

sig/helpers.rbs
```diff
@@ -0,0 +1,35 @@
...
+  #
+  def to_json: (String|Hash[Symbol, String] body) -> String
+
+  # Converts each key of a hash to symbols
+  #
+  def symbolize_hash: (Hash[Symbol|String, String] hash) -> Hash[Symbol, String]
```
DevinCodes, line 8: Do we always force the `Hash[Symbol|String, String]` type? I think it's safer if we opt for `Hash[Symbol|String, untyped]`, so that we can convert nested hashes as well, for example. Wdyt?

sig/iterators/base_iterator.rbs
```diff
@@ -0,0 +1,15 @@
...
+    #TODO: double check Enumerable
+    include Enumerable[Hash[Symbol, String|[String]], void]
+
+    # TODO: add Transporter
+    attr_reader transporter: untyped
+    attr_reader index_name: String
+    attr_reader opts: Hash[Symbol, String|[String]]
+
+    # TODO: add Transporter
+    def initialize: (untyped transporter, String index_name, Hash[Symbol, String|[String]] opts) -> bool
```
DevinCodes, line 13: See https://github.com/algolia/algoliasearch-client-ruby/pull/406/#discussion_r484905987 for an explanation :)
```suggestion
    def initialize: (untyped transporter, String index_name, Hash[Symbol, String|[String]] opts) -> void
```

sig/iterators/paginator_iterator.rbs
```diff
@@ -0,0 +1,10 @@
+module Algolia
+  class PaginatorIterator < BaseIterator
+
+    def initialize: (untyped transporter, String index_name, Hash[Symbol, String|[String]] opts) -> bool
```
DevinCodes, line 4:
```suggestion
    # TODO: type transporter
    def initialize: (untyped transporter, String index_name, Hash[Symbol, String|[String]] opts) -> void
```

DevinCodes (changes requested):

sig/helpers.rbs
chloelbn: completely agree

sig/helpers.rbs
chloelbn: indeed we're not sure what's the second parameter will be

sig/helpers.rbs
```diff
@@ -0,0 +1,12 @@
+module Helpers
+
+  def to_json: (String|Hash[Symbol, String] body) -> String
+  def symbolize_hash: (Hash[Symbol|String, untyped] hash) -> Hash[Symbol, String]
```
DevinCodes, line 4:
```suggestion
  def symbolize_hash: (Hash[Symbol|String, untyped] hash) -> Hash[Symbol, untyped]
```

DevinCodes (changes requested):

DevinCodes (approved):

PR #403, opened 2020-09-07
Feat/add steep transport

chloelbn:
Adds type checking of Transport layer
Depends on https://github.com/algolia/algoliasearch-client-ruby/pull/401

### DIFF
--- a/lib/algolia/transport/request_options.rb
+++ b/lib/algolia/transport/request_options.rb
@@ -34,7 +34,7 @@ module Algolia
       def add_headers(opts = {})
         unless opts[:headers].nil?
           opts[:headers].each do |opt, value|
-            @headers[opt.to_sym] = value
+            @headers[opt.to_s] = value
           end
           opts.delete(:headers)
         end
--- a/lib/algolia/transport/transport.rb
+++ b/lib/algolia/transport/transport.rb
@@ -132,10 +132,7 @@ module Algolia
       # @return [Hash] merged headers
       #
       def generate_headers(request_options)
-        headers                                                = {}
-        extra_headers                                          = request_options.headers
-        @config.default_headers.each { |key, val| headers[key] = val }
-        extra_headers.each { |key, val| headers[key]           = val }
+        headers = @config.default_headers.merge(request_options.headers)
         if request_options.compression_type == Defaults::GZIP_ENCODING
           headers['Accept-Encoding']  = Defaults::GZIP_ENCODING
         end
--- a/lib/algolia/transport/transport.rb
+++ b/lib/algolia/transport/transport.rb
@@ -132,10 +132,10 @@ module Algolia
       # @return [Hash] merged headers
       #
       def generate_headers(request_options)
-        headers                                                     = {}
-        extra_headers                                               = request_options.headers || {}
-        @config.default_headers.each { |key, val| headers[key.to_s] = val }
-        extra_headers.each { |key, val| headers[key.to_s]           = val }
+        headers                                                = {}
+        extra_headers                                          = request_options.headers
+        @config.default_headers.each { |key, val| headers[key] = val }
+        extra_headers.each { |key, val| headers[key]           = val }
         if request_options.compression_type == Defaults::GZIP_ENCODING
           headers['Accept-Encoding']  = Defaults::GZIP_ENCODING
         end
--- a/lib/algolia/transport/transport.rb
+++ b/lib/algolia/transport/transport.rb
@@ -127,11 +127,11 @@ module Algolia
 
       # Generates headers from config headers and optional parameters
       #
-      # @option options [String] :headers
+      # @param request_options [RequestOptions]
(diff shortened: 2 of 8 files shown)


lib/algolia/transport/request_options.rb
```diff
@@ -34,7 +34,7 @@ def create(opts = {})
       def add_headers(opts = {})
         unless opts[:headers].nil?
           opts[:headers].each do |opt, value|
-            @headers[opt.to_sym] = value
+            @headers[opt.to_s] = value
```
DevinCodes, line 37: Why do we cast to a string instead of a symbol?

lib/algolia/transport/transport.rb
```diff
@@ -127,15 +127,15 @@ def build_body(body, request_options, method)
...
-      def generate_headers(request_options = {})
-        headers                                                     = {}
-        extra_headers                                               = request_options.headers || {}
-        @config.default_headers.each { |key, val| headers[key.to_s] = val }
-        extra_headers.each { |key, val| headers[key.to_s]           = val }
+      def generate_headers(request_options)
+        headers                                                = {}
+        extra_headers                                          = request_options.headers
+        @config.default_headers.each { |key, val| headers[key] = val }
+        extra_headers.each { |key, val| headers[key]           = val }
```
DevinCodes, line 138:
```suggestion
        headers = @config.default_headers.merge(request_options.headers)
```

sig/transport/request_options.rbs
```diff
@@ -0,0 +1,54 @@
...
+  module Transport
+    class RequestOptions
+      attr_accessor headers: Hash[String, String]
+      attr_accessor params: Hash[Symbol, String]
+      attr_accessor data: Hash[Symbol, String]
+      attr_accessor timeout: nil|Integer
+      attr_accessor connect_timeout: nil|Integer
+      attr_accessor compression_type: String
+
+      def initialize: (Search::Config config) -> bool
```
DevinCodes, line 11:
```suggestion
      def initialize: (Search::Config config) -> void
```

sig/transport/request_options.rbs
```diff
@@ -0,0 +1,54 @@
...
+
+      # Create and format headers and params from request options
```
DevinCodes, line 13: As with your other PR, I think we can remove the comments in `.rbs` files 🙂

sig/transport/stateful_host.rbs
```diff
@@ -0,0 +1,16 @@
...
+    class StatefulHost
+      include CallType
+      attr_reader url: String
+      attr_reader protocol: String
+      attr_reader accept: String
+      attr_accessor last_use: Time
+      attr_accessor retry_count: Integer
+      attr_accessor up: bool
+
+      def initialize: (String url, ?::Hash[Symbol, String|[String]] opts) -> bool
```
DevinCodes, line 13:
```suggestion
      def initialize: (String url, ?::Hash[Symbol, String|[String]] opts) -> void
```

sig/transport/transport.rbs
```diff
@@ -0,0 +1,20 @@
+module Algolia
+  module Transport
+    class Transport
+      include RetryOutcomeType
+      include CallType
+      #TODO include Helpers
+      include Helpers
+
+      def initialize: (Search::Config config, Object requester) -> bool
```
DevinCodes, line 9:
```suggestion
      def initialize: (Search::Config config, Object requester) -> void
```

DevinCodes (changes requested):

lib/algolia/transport/request_options.rb
chloelbn: The format of the headers will always be string anyway, because of headers like 'X-Algolia-XXX'

DevinCodes: If it's always a string, we don't need to add the `.to_s` right? 🙂

chloelbn: I meant they're supposed to be always string, so it's a way to ensure it's the case

sig/transport/request_options.rbs
```diff
@@ -0,0 +1,21 @@
+module Algolia
+  module Transport
+    class RequestOptions
+      attr_accessor headers: Hash[String, String]
+      attr_accessor params: Hash[Symbol, String]
```
DevinCodes, line 5: Can't the param values be a `bool` or `Integer` too?

sig/transport/transport.rbs
```diff
@@ -0,0 +1,20 @@
+module Algolia
+  module Transport
+    class Transport
+      include RetryOutcomeType
+      include CallType
+      #TODO include Helpers
```
DevinCodes, line 6:
```suggestion
```

DevinCodes (changes requested):

DevinCodes (approved):

PR #405, opened 2020-09-07
ci(cts): fix flaky test, handle 429 HTTP error for set_personalization_strategy

### LINKED ISSUE #397: test: prevent error in CTS personalization test on 429 HTTP code
The personalization API is now limiting the number of
`set_personalization_strategy()` successful calls to 15 per day. Because of
that, our integration tests against the real API is not returning the following
error very often on our testing application:

```json
{
    "status": 429,
    "message": "Number of strategy saves exceeded for the day"
}
```

Because of that, if the CTS is implemented in the client, we need to perform
the `set_personalization_strategy()` call in our integration test like before
but to prevent the test from failing if the 429 error is returned (considered
it as a "success" and moving on with the `get_personalization_strategy()` call).

### LINKED ISSUE #402: Flake test: test_api_keys
```
  1) Error:
API keys#test_api_keys:
NoMethodError: undefined method `<=' for {:maxHitsPerQuery=>42}:Hash
    /home/circleci/project/lib/algolia/responses/update_api_key_response.rb:26:in `wait'
    /home/circleci/project/lib/algolia/search_client.rb:336:in `update_api_key!'
    /home/circleci/project/test/algolia/integration/search_client_test.rb:224:in `test_api_keys'
```

DevinCodes:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| Related Issue     | Fix #397 
| Related Issue     | Fix #402  
| Need Doc update   | no

## Describe your change
This PR adds a `rescue` block around the `set_personalization_strategy` method to ignore HTTP errors with a status code of `429`. It also ensures we do a subset check on a `Set` instead of a `Hash`, so that the code works for Ruby 2.2 as well.

* The endpoint that `set_personalization_strategy` targets only allows 15 updates per day to the Personalization strategies. This means that if we run more integrations tests than these 15, the test suite will fail.
* Flaky/failing test in `test_api_keys`

### DIFF
--- a/lib/algolia/responses/update_api_key_response.rb
+++ b/lib/algolia/responses/update_api_key_response.rb
@@ -24,10 +24,9 @@ module Algolia
         begin
           api_key = @client.get_api_key(@raw_response[:key], opts)
           @done   = @request_options <= api_key
-        rescue StandardError => e
-          if e.code != 404
-            raise e
-          end
+        rescue Algolia::AlgoliaError => e
+          raise e unless e.code == 404
+
           retries_count    += 1
           time_before_retry = retries_count * Defaults::WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY
           sleep(time_before_retry / 1000)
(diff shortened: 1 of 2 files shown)


DevinCodes:
Closed in favor of #400

PR #406, opened 2020-09-08
Feat/add steep http

DevinCodes:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes

## Describe your change
This adds `.rbs` files for all classes in `lib/algolia/http/`, and some which were needed to type these.

### DIFF
--- a/sig/http/http_requester.rbs
+++ b/sig/http/http_requester.rbs
@@ -1,11 +1,11 @@
 module Algolia
   module Http
     class HttpRequester
-      attr_accessor adapter: untyped
+      attr_accessor adapter: String
 
-      attr_accessor logger: untyped
+      attr_accessor logger: Logger
 
-      def initialize: (untyped adapter, untyped logger) -> untyped
+      def initialize: (String adapter, Logger logger) -> void
 
       def send_request: (Algolia::Transport::StatefulHost host, Symbol method, String path, Hash[Symbol | String, untyped] body, Hash[String | Symbol, String | bool] headers, Integer timeout, Integer connect_timeout) -> Algolia::Http::Response
 
--- a/sig/http/http_requester.rbs
+++ b/sig/http/http_requester.rbs
@@ -1,0 +1,17 @@
+module Algolia
+  module Http
+    class HttpRequester
+      attr_accessor adapter: untyped
+
+      attr_accessor logger: untyped
+
+      def initialize: (untyped adapter, untyped logger) -> untyped
+
+      def send_request: (Algolia::Transport::StatefulHost host, Symbol method, String path, Hash[Symbol | String, untyped] body, Hash[String | Symbol, String | bool] headers, Integer timeout, Integer connect_timeout) -> Algolia::Http::Response
+
+      def connection: (Algolia::Transport::StatefulHost host) -> _Connection
+
+      def build_url: (Algolia::Transport::StatefulHost host) -> String
+    end
+  end
+end
--- a/sig/http/http_requester.rbs
+++ b/sig/http/http_requester.rbs
@@ -7,7 +7,7 @@ module Algolia
 
       def initialize: (String adapter, Logger logger) -> void
 
-      def send_request: (Algolia::Transport::StatefulHost host, Symbol method, String path, Hash[Symbol | String, untyped] body, Hash[String | Symbol, String | bool] headers, Integer timeout, Integer connect_timeout) -> Algolia::Http::Response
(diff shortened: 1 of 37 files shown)


sig/http/response.rbs
```diff
@@ -0,0 +1,26 @@
+module Algolia
+  module Http
+    class Response
+      attr_reader status: untyped
```
chloelbn, line 4: you can type all those attributes too

sig/http/http_requester.rbs
```diff
@@ -0,0 +1,17 @@
+module Algolia
+  module Http
+    class HttpRequester
+      attr_accessor adapter: untyped
+
+      attr_accessor logger: untyped
+
+      def initialize: (untyped adapter, untyped logger) -> untyped
```
chloelbn, line 8: `initialize` seems to return a `bool`

sig/http/http_requester.rbs
```diff
@@ -0,0 +1,17 @@
+module Algolia
+  module Http
+    class HttpRequester
+      attr_accessor adapter: untyped
```
chloelbn, line 4: you can type those as well

sig/http/response.rbs
```diff
@@ -0,0 +1,26 @@
...
+      attr_reader network_failure: untyped
+
+      #
+      # @option status    [String]  Response status
+      # @option body    [String]  Response body
+      # @option error    [String]  Response error or caught error
+      # @option headers    [String]  Response headers
+      # @option has_timed_out    [String]  If the request has timed out
+      #
+      def initialize: (?::Hash[Symbol, String | bool] opts) -> untyped
```
chloelbn, line 23: For readability I'd be in favor of deleting the spaces around the `|`, just for the RBS definition. I find that overall easier to read. What are your thoughts? I'd also be in favor of removing the dockblocks are they're not useful in `.rbs` files, and deleting the extra lines between classes variables etc (leaving only one between the variables and the methods)

sig/interfaces/_connection.rbs
```diff
@@ -0,0 +1,16 @@
+interface _Connection
+
+  def options: () -> _ConnectionOptions
+
+  def run_request: (untyped method, untyped path, untyped body, untyped headers) -> untyped
```
chloelbn, line 5:
```suggestion
  def run_request: (Symbol method, String path, Hash[Symbol, String|[String]] body, Hash[String, String] headers) -> Http::Response
```

sig/interfaces/_connection.rbs
```diff
@@ -0,0 +1,16 @@
...
+end
+
+interface _ConnectionOptions
+
+  def timeout: () -> Integer
+  def timeout=: (Integer timeout) -> Integer
+  def open_timeout: () -> Integer
+  def open_timeout=: (Integer timeout) -> Integer
+
+end
```
chloelbn, line 16: Missing extra line and the EOF

sig/transport/request_options.rbs
```diff
@@ -0,0 +1,33 @@
...
+
+      attr_accessor data: Hash[Symbol|String, untyped]
+
+      attr_accessor timeout: Integer
+
+      attr_accessor connect_timeout: Integer
+
+      attr_accessor compression_type: String
+
+      def initialize: (Algolia::Search::Config config) -> void
```
chloelbn, line 16: I don't think it's necessary to prefix with `Algolia::`

chloelbn (changes requested):

sig/http/response.rbs
DevinCodes: Yes, agreed!

sig/http/http_requester.rbs
DevinCodes: While that's true, we should use `void` with the initialize method; this indicates that the return value should not be used (which is the case with every `initialize` method, since we never call it directly. Updating it in all files 👍

sig/transport/request_options.rbs
DevinCodes: It's not, indeed. However, to prevent any namespace clashes, I personally write out the full namespace all the time. I believe other libraries (like Elastic's Ruby client) do the same thing. Wdyt, should we leave it, or remove it?

chloelbn: okay let's keep it!

sig/http/http_requester.rbs
chloelbn: oooh interesting! thanks!

chloelbn (approved):

chloelbn (approved):

PR #408, opened 2020-09-08
feat(find_objects): use lambda instead of block

### LINKED ISSUE #407: Fix: use lambda in find_objects method
The `find_object` method is used to find an object in the index using a function as a parameter. We decided to use a block at the moment, but it might be more idiomatic to use a lambda

DevinCodes:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #407 

## Describe your change
This PR introduces the use of lambda's in the `find_object` method instead of using blocks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,8 +110,8 @@ workflows:
       - test_ruby:
           matrix:
             parameters:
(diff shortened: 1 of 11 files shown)


chloelbn (approved):

chloelbn:
@DevinCodes could you rebase please?

PR #416, opened 2020-09-22
chore(tests): update restore api key test

### LINKED ISSUE #395: Implement retry until success for restore_api_key() call in CTS test
Because of the way Algolia is designed, performing a `restoreApiKey` may not
succeed because the previous `deleteApiKey` takes some time to get propagated
to all Algolia servers, giving us the following response: `[404] Key already
exists`.

To prevent that, we need to change the `testApiKeys` CTS test, if implemented,
to retry even on `[404] Key already exists` response.

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #395
| Need Doc update   | no

### DIFF
--- a/test/algolia/integration/search_client_test.rb
+++ b/test/algolia/integration/search_client_test.rb
@@ -233,7 +233,16 @@ class SearchClientTest < BaseTest
 
       assert_equal 'Key does not exist', exception.message
 
-      @@search_client.restore_api_key!(@api_key[:value])
+      loop do
+        begin
+          @@search_client.restore_api_key!(@api_key[:value])
+          break
+        rescue Algolia::AlgoliaHttpError => e
+          if e.code != 404
+            raise StandardError
+          end
+        end
+      end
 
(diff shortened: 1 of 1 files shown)


test/algolia/integration/search_client_test.rb
```diff
@@ -233,7 +233,15 @@ def test_api_keys
 
       assert_equal 'Key does not exist', exception.message
 
-      @@search_client.restore_api_key!(@api_key[:value])
+      loop do
+        begin
+          return @@search_client.restore_api_key!(@api_key[:value])
```
DevinCodes, line 238: I think this will abort the tests that are below this loop, shouldn't we use `break`?

Like this:
```
begin
  @@search_client.restore_api_key!(@api_key[:value])
  break
rescue ...
```

DevinCodes (changes requested):

DevinCodes (approved):

PR #420, opened 2020-09-24
feat(webmock): add webmock

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     

## Describe your change

Adds back the Webmock class

### DIFF
--- a/lib/algolia.rb
+++ b/lib/algolia.rb
@@ -36,7 +36,6 @@ require 'algolia/recommendation_client'
 require 'algolia/error'
 require 'algolia/search_index'
 require 'algolia/logger_helper'
-require 'algolia/webmock'
 
 # Algolia module
 module Algolia
--- a/lib/algolia.rb
+++ b/lib/algolia.rb
@@ -36,6 +36,7 @@ require 'algolia/recommendation_client'
 require 'algolia/error'
 require 'algolia/search_index'
 require 'algolia/logger_helper'
+require 'algolia/webmock'
 
 # Algolia module
 module Algolia
--- a/lib/algolia/webmock.rb
+++ b/lib/algolia/webmock.rb
@@ -1,0 +1,102 @@
+begin
+  require 'webmock'
+rescue LoadError
+  puts 'WebMock was not found, please add "gem \'webmock\'" to your Gemfile.'
+  exit 1
+end
+
+module Algolia
+  class WebMock
+    def self.mock!
+      # list indexes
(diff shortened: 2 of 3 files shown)


lib/algolia/webmock.rb
```diff
@@ -0,0 +1,102 @@
...
+rescue LoadError
+  puts 'WebMock was not found, please add "gem \'webmock\'" to your Gemfile.'
+  exit 1
+end
+
+module Algolia
+  class WebMock
+    def self.mock!
+      # list indexes
+      ::WebMock.stub_request(:get, %r{.*\.algolia(net\.com|\.net)/1/indexes}).to_return(body: '{ "items": [] }')
```
DevinCodes, line 12: Personally, I find this quite hard to read (because of the repeated `%r{.*\.algolia(net\.com|\.net)/1/indexes}`) mostly. WDYT of refactoring it into a `mock_endpoint` method, so that we can do this:

```ruby
mock_endpoint(:get, '/1/indexes', { body: '{ "items": [] }' })
```

WDYT? 🙂

DevinCodes (commented):
It looks fine to me, but I left a question/comment 🙂  Let me know what you think!

lib/algolia/webmock.rb
chloelbn: I think it's indeed hard to read but at the same time it's not really supposed to be read, just used as an helper to stub calls to the API. Moreover it might be more helpful to read the mock as-is than having it refactored. I would be in favor of keeping it like this

DevinCodes: I'm not sure I agree 😅  I think it should be readable, because it's code that we (and other contributors) have to maintain, even if we don't have to touch it often. If we add a new domain (e.g. `algolia.dev` or something similar) it means we have to refactor the whole class, instead of just a single line of code. I think refactoring this will make contributing easier in the future, so I would be in favor of changing it 🙂

chloelbn: It's true that it's better for refactoring 👍 I'll make the changes

lib/algolia.rb
```diff
@@ -36,6 +36,7 @@
 require 'algolia/error'
 require 'algolia/search_index'
 require 'algolia/logger_helper'
+require 'algolia/webmock'
```
DevinCodes, line 39: I we add this here, we require it by default every time, instead of letting the user decide when to include it 🙂 
```suggestion
```

DevinCodes (changes requested):
One small change, other than that it looks good to me! 🚀

PR #425, opened 2020-10-06
chore(index): renames index_name to name

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no

## Describe your change
Rename `index_name` in the `Index` class as it is redundant and reduce upgrade on the Rails side

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -184,3 +184,6 @@ Style/HashTransformValues:
 
 Style/HashEachMethods:
   Enabled: true
+
+Style/RedundantBegin:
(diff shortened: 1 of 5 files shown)


DevinCodes (changes requested):
Does this require any doc updates too? (I don't think so, but we have to check it to be sure)

Also, can you fix the failing tests, please? 🙂

chloelbn:
@DevinCodes checked the doc, it doesn't need update!

DevinCodes (approved):

PR #432, opened 2020-11-02
chore: update faraday dependency range

### LINKED ISSUE #431: Allow usage with 1.x versions of the Faraday gem
- Algolia Client Version: 2.0.0
- Language Version: 2.6

### Description

We use the [Sentry gem](https://github.com/getsentry/sentry-ruby) which depends on version 1.x of the Faraday gem. Because of this it looks like we can't upgrade from v1 of this gem to v2, due to this gem depending on version 0.x of Faraday. Would it be possible to allow 1.x versions of Faraday?

### Steps To Reproduce

- Run `bundle init` in a new directory
- Run `bundle add sentry-raven` to add the Sentry gem
- Run `bundle add algolia` to add this gem. Bundler raises this error:
```
Bundler could not find compatible versions for gem "faraday":
  In snapshot (Gemfile.lock):
    faraday (= 1.1.0)

  In Gemfile:
    algolia was resolved to 2.0.0, which depends on
      faraday (~> 0.15)

    sentry-raven was resolved to 3.1.1, which depends on
      faraday (>= 1.0)
```

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no   
| Related Issue     | Fix #431 

## Describe your change

Updates the dependency range of `faraday` gem to allow versions > 1.0

### DIFF
--- a/algolia.gemspec
+++ b/algolia.gemspec
@@ -35,7 +35,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rubocop', '<= 0.82.0'
 
-  spec.add_dependency 'faraday', '~> 0.15'
+  spec.add_dependency 'faraday', ['>= 0.15', '< 2.0']
   spec.add_dependency 'multi_json', '~> 1.0'
   spec.add_dependency 'net-http-persistent'
 


algolia.gemspec
```diff
@@ -35,7 +35,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rubocop', '<= 0.82.0'
 
-  spec.add_dependency 'faraday', '~> 0.15'
+  spec.add_dependency 'faraday', ['>= 0.15', '< 2.0']
```
DevinCodes, line 38: Does version `1.x` have backwards compatibility for version `0.x`? 🤔

DevinCodes (changes requested):

algolia.gemspec
chloelbn: it doesn't break anything we're using!

DevinCodes: How did you test that? 🙂

DevinCodes: Okay, great! 😄  Thank you for the explanation 🙂

DevinCodes (approved):

rsl:
curious about the reason for not allowing faraday 2.x+?

DevinCodes:
Hey @rsl ,

As far as I know there's no real reason behind it other than it's not tested. Feel free to open a PR to allow v2 of Faraday as well, and we can take a look 😄

rsl:
@DevinCodes heh i was planning on doing that today. thanks for the thumbs up.

PR #445, opened 2021-01-15
Feat: custom dictionaries

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes/no    
| BC breaks?        | no     
| Related Issue     | Fix #...  
| Need Doc update   | yes/no


lib/algolia/search_client.rb
```diff
@@ -594,12 +594,182 @@ def pending_mappings?(opts = {})
...
+      # @param clear_existing_dictionary_entries [Boolean] delete the previous custom dictionary entries. default: false
+      # @param opts [Hash] contains extra parameters to send with your query
+      #
+      # @return DictionaryResponse
+      #
+      def save_dictionary_entries(dictionary, dictionary_entries, clear_existing_dictionary_entries = false, opts = {})
+        response = @transporter.write(
+          :POST,
+          path_encode('/1/dictionaries/%s/batch', dictionary),
+          { clearExistingDictionaryEntries: clear_existing_dictionary_entries, requests: chunk('addEntry', dictionary_entries) },
```
Haroenv, line 618: I don't know how chunk works, but will this result in an object like this? `{ type: 'addEntry', body: dictionary_entries[i] }`?

lib/algolia/search_client.rb
```diff
@@ -594,12 +594,182 @@ def pending_mappings?(opts = {})
...
+      end
+
+      # Save entries for a given dictionary and wait for the task to finish
+      #
+      # @param dictionary [String] dictionary name. Can be either 'stopwords', 'plurals' or 'compounds'
+      # @param dictionary_entries [Array<Hash>] array of dictionary entries
+      # @param clear_existing_dictionary_entries [Boolean] delete the previous custom dictionary entries. default: false
+      # @param opts [Hash] contains extra parameters to send with your query
+      #
+      def save_dictionary_entries!(dictionary, dictionary_entries, clear_existing_dictionary_entries = false, opts = {})
```
Haroenv, line 632: if we have a separate method for `replace`, maybe the `clear_existing_dictionary_entries` boolean shouldn't be exposed in save

lib/algolia/search_index.rb
```diff
@@ -1045,47 +1045,6 @@ def check_array(object)
         raise AlgoliaError, 'argument must be an array of objects' unless object.is_a?(Array)
       end
 
-      # Check the passed object
-      #
-      # @param object [Object]
-      # @param in_array [Boolean] whether the object is an array or not
-      #
-      def check_object(object, in_array = false)
```
Haroenv, line 1053: not sure what this is, but the change doesn't look related

Haroenv (commented):
👌

lib/algolia/search_client.rb
chloelbn: the object will look like this 
`{ action: 'addEntry', body: dictionary_entries[i] }`

lib/algolia/search_index.rb
chloelbn: it's kind of related as part of those methods are used in the search client as well, and started to be needed for this feature. But I can make a different PR

Haroenv: yes, I think separate PR will be more straightforward :)

chloelbn: it's done and merged!

lib/algolia/responses/dictionary_response.rb
```diff
@@ -0,0 +1,37 @@
...
+      retries_count = 1
+
+      until @done
+        res               = @client.custom_request({}, path_encode('/1/task/%s', @raw_response[:taskID]), :GET, READ)
+        status            = get_option(res, 'status')
+        if status == 'published'
+          @done = true
+        end
+        retries_count    += 1
+        time_before_retry = retries_count * Defaults::WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY
```
DevinCodes, line 30: This increases the delay with each loop that we do: is that intended? If so, why don't we keep the time between retries the same?

 E.g. if the first loop we wait 1 second, we wait 5 seconds at the fifth loop. Why not keep it 1 second all the time?

lib/algolia/search_client.rb
```diff
@@ -594,12 +594,186 @@ def pending_mappings?(opts = {})
...
       #
+      # @param dictionary [String] dictionary name. Can be either 'stopwords', 'plurals' or 'compounds'
+      # @param object_ids [Array<Hash>] array of object ids
+      # @param opts [Hash] contains extra parameters to send with your query
+      #
+      def delete_dictionary_entries!(dictionary, object_ids, opts = {})
+        request  = []
+        object_ids.map do |object_id|
+          request.push({ objectID: object_id })
+        end
```
DevinCodes, line 696: Can you move this logic into the `delete_dictionary_entries` method? Right now, the usage of `delete_dictionary_entries!` and `delete_dictionary_entries` differs (one takes an array of objectIDs, the other an array of formatted hashes). Moving this logic into `delete_dictionary_entries` allows to reuse the same API for both methods.

test/algolia/integration/search_client_test.rb
```diff
@@ -367,5 +368,96 @@ def test_expired_secured_api_keys
...
+        assert_equal 1, stopwords[:nbHits]
+        assert_equal stopwords[:hits][0][:objectID], entry[:objectID]
+        assert_equal stopwords[:hits][0][:word], entry[:word]
+
+        @client.delete_dictionary_entries!('stopwords', [entry_id])
+        assert_equal 0, @client.search_dictionary_entries('stopwords', entry_id)[:nbHits]
+
+        old_dictionary_state   = @client.search_dictionary_entries('stopwords', '')
+        old_dictionary_entries = old_dictionary_state[:hits].map do |hit|
+          hit.reject! { |key| key == :type }
```
DevinCodes, line 398:
```suggestion
          hit.reject { |key| key == :type }
```

the [`reject!` method](https://ruby-doc.org/core-2.6/Hash.html#reject-21-method) returns `nil` if the record hasn't changed, which may cause problems in your tests. Since your using `map` here, you don't need to use the `reject!`.

DevinCodes (changes requested):
Just some small feedback, but it looks great overall 💯

lib/algolia/responses/dictionary_response.rb
chloelbn: we wait a bit longer in case there's latency

lib/algolia/search_client.rb
```diff
@@ -673,10 +673,14 @@ def replace_dictionary_entries!(dictionary, dictionary_entries, opts = {})
       # @return DictionaryResponse
       #
       def delete_dictionary_entries(dictionary, object_ids, opts = {})
+        request  = []
+        object_ids.map do |object_id|
+          request.push({ objectID: object_id })
+        end
```
DevinCodes, line 679: Since you're using `map`, you can assign the result to `request` directly.
```suggestion
        request = object_ids.map do |object_id|
          { objectID: object_id }
        end
```

lib/algolia/responses/dictionary_response.rb
DevinCodes: The latency is handled by the `@client`, right? Basically, with each loop we're waiting longer and longer to check the status of a task, while the chance of it being done is bigger the longer we wait. So I would actually expect that the time between requests stays the same or even reduces, if that makes sense. WDYT?

DevinCodes (changes requested):

lib/algolia/search_client.rb
```diff
@@ -594,12 +594,186 @@ def pending_mappings?(opts = {})
...
+        DictionaryResponse.new(self, response)
+      end
+
+      # Save entries for a given dictionary and wait for the task to finish
+      #
+      # @param dictionary [String] dictionary name. Can be either 'stopwords', 'plurals' or 'compounds'
+      # @param dictionary_entries [Array<Hash>] array of dictionary entries
+      # @param opts [Hash] contains extra parameters to send with your query
+      #
+      def save_dictionary_entries!(dictionary, dictionary_entries, opts = {})
```
VladislavFitz, line 630: Just curious to know: why do you have two declarations for each method? Sync & async version?

chloelbn: the `!` is the equivalent of the method chained with a `wait()`, it was done like this historically and was an appreciated feature so we kept it that way

VladislavFitz: I see, thanks 🙂

DevinCodes (approved):
Looks good to me! 💯

VladislavFitz (approved):
👍

Aallam (approved):

PR #449, opened 2021-02-10
fix(requester): update Unreachable Hosts message

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     

## Describe your change

Update the Unreachable Hosts error message to better reflect what was the last failure.

Note: as in debug mode most of the errors [are already logged](https://github.com/algolia/algoliasearch-client-ruby/blob/master/lib/algolia/http/http_requester.rb#L38), I didn't opt for the error array solution.

### DIFF
--- a/lib/algolia/transport/transport.rb
+++ b/lib/algolia/transport/transport.rb
@@ -47,6 +47,7 @@ module Algolia
       # @return [Response] response of the request
       #
       def request(call_type, method, path, body = {}, opts = {})
+        retry_error = nil
         @retry_strategy.get_tryable_hosts(call_type).each do |host|
           opts[:timeout]         ||= get_timeout(call_type) * (host.retry_count + 1)
           opts[:connect_timeout] ||= @config.connect_timeout * (host.retry_count + 1)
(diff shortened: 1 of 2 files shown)


DevinCodes (approved):
LGTM!

Unrelated, but I noticed it from the code you linked to in your comment. I see we use the following code a lot throughout the client:

```ruby
if ENV['ALGOLIA_DEBUG']
  @logger.info("debug stuff here")
end
```

Can we wrap this into a method, to clean everything up a bit? It saves a lot of duplicated code. WDYT?

Haroenv (approved):

Haroenv:
technically a change in error message is a breaking change though, people could catch only if the message is the previous version (see spacebar heater https://xkcd.com/1172/)

chloelbn:
@DevinCodes yes! I'll do it in another PR
@Haroenv indeed, moreover I changed the tests so I've should known better. What if we add that information in the debug mode instead then?

chloelbn:
I'm closing this issue as changing the error message is a BC break and all the necessary information are already in the logger.

PR #451, opened 2021-03-08
fix(set-settings): forwardToReplicas param

### LINKED ISSUE #450: index.set_settings => { forwardToReplicas: true } as options raises an error
- Algolia Client Version: 2.0.4 (latest)
- Language Version: Ruby

### Description

When I try to set settings and forward to replicas it doesn't work.

```ruby
index.set_settings(settings_hash, { forwardToReplicas: true })
```

### Steps To Reproduce

Simply past the example from documentation https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/how-to/set-settings-and-forward-to-replicas/?client=ruby#using-the-api

```ruby
index.set_settings({ searchableAttributes: ['name', 'description'] }, { forwardToReplicas: true })
```

The error:

```
         4: from /Users/skelz0r/.rvm/gems/ruby-2.6.6/gems/algolia-2.0.4/lib/algolia/search_index.rb:990:in `set_settings'
         3: from /Users/skelz0r/.rvm/gems/ruby-2.6.6/gems/algolia-2.0.4/lib/algolia/transport/transport.rb:38:in `write'
         2: from /Users/skelz0r/.rvm/gems/ruby-2.6.6/gems/algolia-2.0.4/lib/algolia/transport/transport.rb:50:in `request'
         1: from /Users/skelz0r/.rvm/gems/ruby-2.6.6/gems/algolia-2.0.4/lib/algolia/transport/transport.rb:50:in `each'
/Users/skelz0r/.rvm/gems/ruby-2.6.6/gems/algolia-2.0.4/lib/algolia/transport/transport.rb:72:in `block in request': Invalid object attributes: forwardToReplicas near line:1 column:128 (Algolia::AlgoliaHttpError)
```

## Inspecting

The `set_settings` method is defined here: https://github.com/algolia/algoliasearch-client-ruby/blob/f7009fc0572f8aa1cd198acd1a2ced8900a2e165/lib/algolia/search_index.rb#L989

When we dig another method (like `save_synonyms`) I noticed that the `forwardToReplicas` is directly build on the `path` parameter ( https://github.com/algolia/algoliasearch-client-ruby/blob/f7009fc0572f8aa1cd198acd1a2ced8900a2e165/lib/algolia/search_index.rb#L669 ) which is not the case for `set_settings`

## Workaround

After some digging in the `Algolia::Transport::Transport` class I managed to find a way to pass my parameter:

```ruby
index.set_settings({ searchableAttributes: ['name', 'description'] }, { params: { forwardToReplicas: true }})
```

But it's not consistent with `save_synonyms` (and perhaps with other methods).

chloelbn:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #450
| Need Doc update   | yes/no

The parameter `forwardToReplicas` should be handled independently in the `set_settings` method

### DIFF
--- a/lib/algolia/search_index.rb
+++ b/lib/algolia/search_index.rb
@@ -987,13 +987,8 @@ module Algolia
       # @return [IndexingResponse]
       #
       def set_settings(settings, opts = {})
-        forward_to_replicas  = false
         request_options      = symbolize_hash(opts)
-
-        if request_options[:forwardToReplicas]
-          forward_to_replicas = true
-          request_options.delete(:forwardToReplicas)
-        end
+        forward_to_replicas  = request_options.delete(:forwardToReplicas) || false
 
         response = @transporter.write(
           :PUT,
--- a/lib/algolia/search_index.rb
+++ b/lib/algolia/search_index.rb
@@ -987,7 +987,20 @@ module Algolia
       # @return [IndexingResponse]
       #
       def set_settings(settings, opts = {})
-        response = @transporter.write(:PUT, path_encode('/1/indexes/%s/settings', @name), settings, opts)
+        forward_to_replicas  = false
+        request_options      = symbolize_hash(opts)
+
+        if request_options[:forwardToReplicas]
+          forward_to_replicas = true
+          request_options.delete(:forwardToReplicas)
+        end
+
+        response = @transporter.write(
+          :PUT,
+          path_encode('/1/indexes/%s/settings', @name) + handle_params({ forwardToReplicas: forward_to_replicas }),
+          settings,
+          request_options
+        )
 
         IndexingResponse.new(self, response)
       end
--- a/test/algolia/integration/search_index_test.rb
+++ b/test/algolia/integration/search_index_test.rb
@@ -274,6 +274,9 @@ class SearchIndexTest < BaseTest
       @index.set_settings!(settings)
 
       assert_equal @index.get_settings, settings
+
+      # check that the forwardToReplicas parameter is passed correctly
+      assert @index.set_settings!(settings, { forwardToReplicas: true })
     end
   end
 


lib/algolia/search_index.rb
```diff
@@ -987,7 +987,20 @@ def get_settings(opts = {})
...
       #
       def set_settings(settings, opts = {})
-        response = @transporter.write(:PUT, path_encode('/1/indexes/%s/settings', @name), settings, opts)
+        forward_to_replicas  = false
+        request_options      = symbolize_hash(opts)
+
+        if request_options[:forwardToReplicas]
+          forward_to_replicas = true
+          request_options.delete(:forwardToReplicas)
+        end
```
DevinCodes, line 996: The [`Hash#delete`](https://apidock.com/ruby/Hash/delete) method returns the value of the corresponding key, or `nil` if not defined. We can use that behavior to shorten the code a lot:
```suggestion
        request_options      = symbolize_hash(opts)
        forward_to_replicas  = request_options.delete(:forwardToReplicas) || false
```

test/algolia/integration/search_index_test.rb
```diff
@@ -274,6 +274,9 @@ def test_settings
       @index.set_settings!(settings)
 
       assert_equal @index.get_settings, settings
+
+      # check that the forwardToReplicas parameter is passed correctly
+      assert @index.set_settings!(settings, { forwardToReplicas: true })
```
DevinCodes, line 279: Does this test break _without_ your changes in `lib/algolia/search_index.rb`?

DevinCodes (changes requested):

test/algolia/integration/search_index_test.rb
chloelbn: indeed! it throws `Algolia::AlgoliaHttpError: Invalid object attributes: forwardToReplicas`

DevinCodes (changes requested):
It looks good to me, but the tests are failing. Can you check why, please?

DevinCodes (approved):

PR #452, opened 2021-03-26
chore(cts): add retry

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     

## Describe your change

Add a retry strategy to prevent flakiness in the CI

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -87,3 +87,15 @@ def rule_without_metadata(rule)
   rule.delete(:_metadata)
   rule
 end
+
+def retry_test(delay = 0.1, max_retries = 30)
+  (1...max_retries).each do |i|
+    begin
+      return yield
+    rescue Algolia::AlgoliaHttpError
+      sleep delay * i
+    end
+  end
+
+  raise StandardError, 'reached the maximum number of retries'
+end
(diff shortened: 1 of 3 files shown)


test/test_helper.rb
```diff
@@ -87,3 +87,15 @@ def rule_without_metadata(rule)
...
+def retry_test(delay = 0.1, max_retries = 30)
+  (1...max_retries).each do |i|
+    begin
+      return yield
+    rescue Algolia::AlgoliaHttpError
+      sleep delay * i
+    end
+  end
+
+  raise StandardError, 'reached the maximum number of retries'
```
DevinCodes, line 100: Can we make a custom error for this instead of using a `StandardError`? Something like a `TooManyRetries` error for example.

DevinCodes (changes requested):

test/test_helper.rb
chloelbn: In this repo we only have two kind of errors, `AlgoliaError` and `AlgoliaHttpError`. I found that it belonged to none of them so I stuck with the standard error

DevinCodes (changes requested):
It looks good to me, can you just have a look why the tests are failing please? 🙂

DevinCodes (approved):

PR #455, opened 2021-05-17
fix: Memoize host based on accept attribute

### LINKED ISSUE #454: Making read-only request breaks subsequent write requests
See https://github.com/algolia/algoliasearch-rails/issues/397 for more background info.

- Algolia Client Version: 2.0.4
- Language Version: ruby 2.7.2

### Description

https://github.com/algolia/algoliasearch-client-ruby/blob/5e7c715a596ac5e01064a16e9f46d3aaa825850b/lib/algolia/http/http_requester.rb#L67-L71

When making a read request the `HttpRequester` memoizes a `@connection` with a read-only host. If afterwards you try to make a write request, then `HttpRequester` will ignore the supplied write host and instead use the memoized `@connection` which was created with the read-only host. This leads to the following error:

`Algolia::AlgoliaHttpError: This host is read-only (ie. this is a DSN), write operations are not accepted`

### Steps To Reproduce

I'm using the `algoliasearch-rails` gem which I think makes a `check_settings` read-request automatically. So any write request fails. Here's a code example using just `algolia-client-ruby`:

```ruby
client = Algolia::Search::Client.create "…", "…"
index = client.init_index("…")

# Make a read request
index.search("a")

# Make a write request
index.save_objects([objectID: 1, name: "Foo"])

# You will get this error
# *** Algolia::AlgoliaHttpError Exception: This host is read-only (ie. this is a DSN), write operations are not accepted
```

### Proposed solution

Make sure the connection is specific to the provided host. Something along these lines:

```ruby
def connection(host)
  @connections[host] ||= Faraday.new(build_url(host)) do |f|
    f.adapter @adapter.to_sym
   end
 end
```

marckohlbrugge:
Before this change a read-only host could get memoized preventing writes in subsequent requests.

Closes #454

| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #454 
| Need Doc update   | no

## Describe your change

Before this change the first used host would get memoized. If this was a read-only host, then future requests would also use this read-only host. Even if this requests would require a write host.

This PR fixes that bug. The existing code comments already referred to a `@connections` variable. So I think this solution might have been implemented before, but got removed for some reason.

I couldn't get the test suite to run, so I suggest doing that before merging this PR. I also suggest adding a spec that fails without this PR. So that this bug doesn't re-occur in the future.

See above.

### DIFF
--- a/lib/algolia/http/http_requester.rb
+++ b/lib/algolia/http/http_requester.rb
@@ -9,8 +9,8 @@ module Algolia
       # @param logger [Object] logger used to log requests. Defaults to Algolia::LoggerHelper
       #
       def initialize(adapter, logger)
-        @adapter    = adapter
-        @logger     = logger
+        @adapter     = adapter
+        @logger      = logger
         @connections = []
       end
 
--- a/lib/algolia/http/http_requester.rb
+++ b/lib/algolia/http/http_requester.rb
@@ -11,7 +11,7 @@ module Algolia
       def initialize(adapter, logger)
         @adapter     = adapter
         @logger      = logger
-        @connections = []
+        @connections = {}
       end
 
       # Sends request to the engine
--- a/lib/algolia/http/http_requester.rb
+++ b/lib/algolia/http/http_requester.rb
@@ -11,7 +11,7 @@ module Algolia
       def initialize(adapter, logger)
         @adapter    = adapter
         @logger     = logger
-        @connection = nil
+        @connections = []
       end
 
       # Sends request to the engine
@@ -65,7 +65,7 @@ module Algolia
       # @return [Faraday::Connection]
       #
       def connection(host)
-        @connection ||= Faraday.new(build_url(host)) do |f|
+        @connections[host.accept] ||= Faraday.new(build_url(host)) do |f|
           f.adapter @adapter.to_sym
         end
       end


lib/algolia/http/http_requester.rb
```diff
@@ -11,7 +11,7 @@ class HttpRequester
       def initialize(adapter, logger)
         @adapter    = adapter
         @logger     = logger
```
DevinCodes, line 13:
```suggestion
        @adapter     = adapter
        @logger      = logger
```

DevinCodes (changes requested):
This should fix the error raised by rubocop. I'm looking into why the test suite doesn't pass.

lib/algolia/http/http_requester.rb
```diff
@@ -9,9 +9,9 @@ class HttpRequester
       # @param logger [Object] logger used to log requests. Defaults to Algolia::LoggerHelper
       #
       def initialize(adapter, logger)
-        @adapter    = adapter
-        @logger     = logger
-        @connection = nil
+        @adapter     = adapter
+        @logger      = logger
+        @connections = []
```
DevinCodes, line 14:
```suggestion
        @connections = {}
```

DevinCodes (changes requested):
While this works with an array for `@connections` (we use numbers for the `READ` and `WRITE` constants), let's use a hash to make it work with strings too if we have to. After this change I'll approve & merge, and release soon after!

DevinCodes (approved):

DevinCodes:
This is released in version 2.1.1 ! https://rubygems.org/gems/algolia/versions/2.1.1

PR #461, opened 2021-07-13
chore: deprecate RecommendationClient in favor of PersonalizationClient

DevinCodes:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change
As per our internal specification, this renames the `RecommendationClient` to `PersonalizationClient`. For backwards compatibility you'll still be able to use the `RecommendationClient` until the next major version.


lib/algolia/personalization_client.rb
```diff
@@ -0,0 +1,60 @@
...
+      # @return self
+      #
+      def self.create(app_id, api_key)
+        config = Personalization::Config.new(application_id: app_id, api_key: api_key)
+        create_with_config(config)
+      end
+
+      # Create a new client providing only an Recommendation::Config object
+      #
+      # @param config [Recommendation::Config]
```
sbellone, line 33:
```suggestion
      # Create a new client providing only an Personalization::Config object
      #
      # @param config [Personalization::Config]
```

lib/algolia/config/recommendation_config.rb
```diff
@@ -1,5 +1,6 @@
 module Algolia
   module Recommendation
+    # <b>DEPRECATED:</b> Please use <tt>Algolia::Personalization::Config</tt> instead.
     class Config < BaseConfig
```
julienbourdeau, line 4: Is there a reason you didn't go for the same solution as the client? extend the Personalization version and remove all content in this class?
```suggestion
    class Config < Algolia::Personalization::Config
```

julienbourdeau (approved):

sbellone (commented):
Just need to rename the called endpoint, otherwise looks good 👌

DevinCodes:
Thank you both! I've updated the PR to address your suggestions 🙂

sbellone (approved):

sbellone (approved):

PR #468, opened 2022-01-10
fix: allow hash in multiple queries

### LINKED ISSUE #462: `#multiple_queries` does not accept a params Hash as documented
### Description
The Ruby API documentation for `#multiple_queries` [shows that the method should accept an array of query objects that look like this](https://www.algolia.com/doc/api-reference/api-methods/multiple-queries/#method-param-queries):

```ruby
{
  indexName: "indexName",
  type: "default", // optional
  query: "search query",
  params: {
    searchParameter1: value1, // one or several search parameters
    searchParameter2: value2
  }
}
```
But, when I try to pass such an object, I get back a 400 error: `Algolia::AlgoliaHttpError (400: Object forbidden at this position near line:1 column:55)`

It seems like what the method is _actually_ expecting is that the params be url form encoded, something like this:

```ruby
{
  indexName: "indexName",
  type: "default", // optional
  query: "search query",
  params: "searchParameter1=value1&searchParameter2=value2"
}
```

[Your tests seem to confirm this](https://github.com/algolia/algoliasearch-client-ruby/blob/5e7c715a596ac5e01064a16e9f46d3aaa825850b/test/algolia/integration/search_client_test.rb#L295)

```ruby
    results = @@search_client.multiple_queries([
        { indexName: index_name1, params: to_query_string({ query: '', hitsPerPage: 2 }) },
        { indexName: index_name2, params: to_query_string({ query: '', hitsPerPage: 2 }) }
      ], { strategy: 'none' })[:results]
```

It would be far more convenient if the method took a params Hash rather than a form encoded string, especially because that's the format that the InstantSearch.js produces--my use case involves proxying InstantSearch.js generated requests through our Ruby backend to enrich results.  I'm not able to just pass the `requests` request body array straight into `#multiple_queries`, but have to map through it and form encode the params first.

### Steps To Reproduce
1. Format an array of queries according to the documentation, where the value of the `params` key is a Hash.
2. Pass that array into `#multiple_queries`.
3. Observe a 400 error.

DevinCodes:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #462 
| Need Doc update   | no

## Describe your change
This updates the client to accept a hash for `params` when performing a `multiple_queries`, rather than only accepting a string.

### DIFF
--- a/lib/algolia/search_client.rb
+++ b/lib/algolia/search_client.rb
@@ -478,6 +478,9 @@ module Algolia
       # @return [Hash]
       #
       def multiple_queries(queries, opts = {})
+        queries.map! do |q|
+          q[:params] = to_query_string(q[:params]) unless q[:params].is_a? String
+        end
         @transporter.read(:POST, '/1/indexes/*/queries', { requests: queries }, opts)
       end
       alias_method :search, :multiple_queries
--- a/lib/algolia/search_client.rb
+++ b/lib/algolia/search_client.rb
@@ -478,8 +478,8 @@ module Algolia
       # @return [Hash]
       #
       def multiple_queries(queries, opts = {})
-        queries.map! do |q|
-          q[:params] = to_query_string(q[:params]) unless q[:params].is_a? String
+        queries.each do |q|
+          q[:params] = to_query_string(q[:params]) unless q[:params].is_a?(String)
         end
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
this is a breaking change for a case that was previously erroring where an actual hash for params was expected to be sent. As the engine doesn't accept that, this works!

PR #471, opened 2022-02-16
chore: update faraday dependency range

daedric:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change
Update Faraday dependency.

Strong dependency on "old" Faraday major.

### DIFF
--- a/algolia.gemspec
+++ b/algolia.gemspec
@@ -35,7 +35,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rubocop', '<= 0.82.0'
 
-  spec.add_dependency 'faraday', ['>= 0.15', '< 2.0']
+  spec.add_dependency 'faraday', ['>= 0.15', '< 3']
+  spec.add_dependency 'faraday-net_http_persistent', ['>= 0.15', '< 3']
   spec.add_dependency 'multi_json', '~> 1.0'
   spec.add_dependency 'net-http-persistent'
 
--- a/lib/algolia/transport/transport.rb
+++ b/lib/algolia/transport/transport.rb
@@ -1,4 +1,6 @@
 require 'faraday'
+# this is the default adapter and it needs to be required to be registered.
+require 'faraday/net_http_persistent'
 
 module Algolia
   module Transport


daedric:
I was not able to run the test however in some private code I was able to use Faraday 2.2 without any issue :)

sfate:
@DevinCodes any news on this one?

algolia.gemspec
```diff
@@ -35,7 +35,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rubocop', '<= 0.82.0'
 
-  spec.add_dependency 'faraday', ['>= 0.15', '< 2.0']
+  spec.add_dependency 'faraday', ['>= 2', '< 3']
```
DevinCodes, line 38: We may want to allow older versions here too. If someone has the faraday dependency in their project this range may prevent them from upgrading. WDYT?
```suggestion
  spec.add_dependency 'faraday', ['>= 0.15', '< 3']
```

algolia.gemspec
```diff
@@ -35,7 +35,8 @@ Gem::Specification.new do |spec|
...
+  spec.add_dependency 'faraday-net_http_persistent', '~> 2.0'
```
DevinCodes, line 39: If we go that route, we should probably have the same range for this dependency
```suggestion
  spec.add_dependency 'faraday-net_http_persistent', ['>= 0.15', '< 3']
```

DevinCodes (changes requested):
Thank you for the PR! I have a remark, let me know what you think 🙂

aaronstillwell:
@DevinCodes @daedric any help needed here to get this merged? I'm unfortunately blocked from using `algoliasearch-rails` until this is resolved - happy to contribute!

algolia.gemspec
daedric: @DevinCodes Faraday 2 is not backward compatible, I cannot use this range (IIRC)

daedric:
Done, I'm sorry I missed the (old) notification :)

DevinCodes:
Quick update; this change seems to require us to drop support for some older Ruby versions, so we're internally discussing how we can address this properly. Will keep you posted!

courtsimas:
👀 

this ruby gem prevents the gem `google-cloud-pubsub`  from resolving due to their `~> 2.0` on faraday. Any update?

aaronstillwell:
While we wait for the team to get back to us on their decision moving forward, we can work against @daedric's fork with:

```ruby
gem 'algolia', git: 'https://github.com/daedric/algoliasearch-client-ruby', ref: '911b338'
```

bvogel:
> Quick update; this change seems to require us to drop support for some older Ruby versions, so we're internally discussing how we can address this properly. Will keep you posted!

I would really move froward and stop supporting ruby versions that are well behind EOL, even 2.6 is now out of support so even maintaining 2.5 would be generous. As for faraday which you rely upon they don't support ruby versions after their EOL and even the support for 0.x was stopped as only the 1.x and 2.x received releases lately. 

I don't understand why algolia is so desperately lagging in version updates as this general approach of "leave no one behind" will not foster a more general consensus for developers to keep upgrading their dependencies and by that keeping their applications unsafe as even security fixes aren't applied bc of "version constraints"

DevinCodes:
Sorry for the delay in coming back to you. We're indeed going to drop support for the older Ruby versions. We're not going to release today given that it's a Friday, but I'm working on this first thing Monday morning (Paris time). I'll drop a note here when a release is made!

DevinCodes (approved):

PR #369, opened 2019-10-04
Modify WebMock sample to testable with rails

kiyohara:
| Q                 | A
| ----------------- | ----------
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | 
| Need Doc update   | yes ( https://www.algolia.com/doc/api-client/guides/ruby/webmock/ ) 

## Describe your change

Update response params about pagination, for testing rails code that include algoliasearch-rails gem.
Modify to include one entry on the page.

Otherwise `page: 0` is useful, maybe.

## What problem is this fixing?

Currently, Model's `algolia_search` method (*1) return empty list (*2).
This response is a little inconvenient for testing search.
This change will become to entry present test assertable.

Yes, I know this file is webmock SAMPLE, this PR priority very low...

(*1) `algolia_search` method is defined in algoliasearch-rails gem.
(*2) The empty list is created via Kaminari::PaginatableArray with only 1 value and page param is 1 (not 0).

PR #472, opened 2022-03-29
fix: API polling flood issue due to msec sleep time mishandling

kiyohara:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | -
| Need Doc update   | no

## Describe your change

There are some `sleep` with `Defaults::WAIT_TASK_DEFAULT_TIME_BEFORE_RETRY / 1000` and the constant variable is `100` as 100 milliseconds.
However, `100 / 1000` is `0` calculated as Integer.
So, I fix it to `100.to_f / 1000` for 100 msec sleeping.

The changes in `lib/algolia/responses/*.rb` are not so important because these are irregular error handling logic and sleep time is increased by retry count.
On the other hand, `lib/algolia/search_index.rb` is important.
It's normal logic that occurs frequently.

Ref. I found out similar logic [here](https://github.com/algolia/algoliasearch-client-ruby/blob/052e96149abb31c67a5ddf3ec2fda9504c3ce8bf/lib/algolia/search_client.rb#L64) that handles the msec as float correctly.

When you use the search method, especially with index settings changes, the client generates API call flood for the status check polling.
This PR reduces the flood, increases the interval of API calls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,3 +62,4 @@ If you were using the v1 and wish to update to v2, please follow our [Upgrade Gu
 ## 📄 License
 
 Algolia Ruby API Client is an open-sourced software licensed under the [MIT license](LICENSE.md).
+
--- a/README.md
+++ b/README.md
@@ -62,4 +62,3 @@ If you were using the v1 and wish to update to v2, please follow our [Upgrade Gu
 ## 📄 License
 
 Algolia Ruby API Client is an open-sourced software licensed under the [MIT license](LICENSE.md).
-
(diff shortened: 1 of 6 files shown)


kiyohara:
Travis can't run any specs because of the same reason as [this](https://github.com/algolia/algoliasearch-client-ruby/pull/369#issuecomment-539957009).
Is there anything I can do?

DevinCodes (approved):
Thank you for so much for this @kiyohara , that's a great catch.

We'll release this on Tuesday (Monday is a day off here in France).

kiyohara:
@DevinCodes I'm really glad I could help!

DevinCodes:
Today has been quite busy, so I have to wait with the release until tomorrow. Sorry for the delay!

kiyohara:
Thanks for letting me know.
Of course, I would be happy to have you proceed at your convenience! :pray:

Also, please check out my other PRs whenever it is convenient for you. I've added tests.
There is no reason to hurry. :bow:

kiyohara:
Thanks for the info. I'm so glad!
We are starting to consider Algolia gem's updating in our project :+1:

PR #473, opened 2022-03-29
fix: API endpoint URL mishandling in get_settings method

kiyohara:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | -
| Need Doc update   | no

## Describe your change

When the `get_settings` method is called in [algoliasearch-rails gem](https://github.com/algolia/algoliasearch-rails/blob/2358e3ca6e3b2242a345b3e2d360e2099fbbc9d6/lib/algoliasearch-rails.rb#L796) with argument `(:getVersion => 1)`, the client send HTTP GET to `/1/indexes/%s/settings?getVersion=2?getVersion=1`
`?getVersion=2?getVersion=1` is invalid params and the API returns the JSON response contain `"version"=>18446744073709551615`.

This PR fix the HTTP GET target URL with some opts.

- When you use `get_settings` method with some opts you may don't get a correct response.
- algoliasearch-rails gem may mis-detect the changes in index settings.

### DIFF
--- a/lib/algolia/search_index.rb
+++ b/lib/algolia/search_index.rb
@@ -977,8 +977,8 @@ module Algolia
         opts_default = {
           getVersion: 2
         }
-        opts     = opts_default.merge(opts)
-        response = @transporter.read(:GET, path_encode('/1/indexes/%s/settings', @name), {}, opts)
+        opts         = opts_default.merge(opts)
+        response     = @transporter.read(:GET, path_encode('/1/indexes/%s/settings', @name), {}, opts)
 
         deserialize_settings(response, @config.symbolize_keys)
       end
--- a/lib/algolia/search_index.rb
+++ b/lib/algolia/search_index.rb
@@ -974,7 +974,11 @@ module Algolia
       # @return [Hash]
       #
       def get_settings(opts = {})
-        response = @transporter.read(:GET, path_encode('/1/indexes/%s/settings', @name) + handle_params({ getVersion: 2 }), {}, opts)
+        opts_default = {
+          getVersion: 2
+        }
+        opts     = opts_default.merge(opts)
+        response = @transporter.read(:GET, path_encode('/1/indexes/%s/settings', @name), {}, opts)
 
         deserialize_settings(response, @config.symbolize_keys)
       end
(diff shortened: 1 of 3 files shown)


DevinCodes (changes requested):
Hi @kiyohara ,

Would you please be able to add a test for this to ensure it works as expected? The code looks good to me, but let's be 100% sure with the tests.
Let me know if you need a hand with that 🙂 

Thank you in advance!

kiyohara:
@DevinCodes Thank you for your review.
I added the version check tests. Would you check it, please?

To be honest, I don't think this test validates changes in this PR strictly.
This PR fixes the URL query building logic for the setting API call, not version value handling.
If the backend server behavior with invalid version query changes, you can't check the URL query building logic correctness by this test.

However, URL checking that requires hooking of API calls is difficult in this test set, it's my opinion.
Also, I have added a backend check test that detects the server behavior changes.
Then this test is suitable at least currently, I think.

What do you think?

DevinCodes (changes requested):
In regards to the tests, maybe we could do something like we do for the tests of the `Recommend` client (https://github.com/algolia/algoliasearch-client-ruby/blob/master/test/algolia/integration/recommend_client_test.rb#L7)

Here, we initialize a search client with a mock requester, and we can test that the mock requester received certain requests. So we could do the check on the proper URL formatting, if that makes sense 😄 

Please let me know if you need a hand with this!

kiyohara:
@DevinCodes Thanks for the advice.
I think I have made the ideal test using the mock integration you described.
May I ask you to check it again?

kiyohara:
@DevinCodes It looks like the test ran correctly thanks to your merging. Thank you so much.

However, It looks like some errors occurred in some tests in all environments.
Given my fixes, it's not immediately clear why these tests are having errors.
Are my fixes adversely affecting the other tests?
Should I attempt to fix them?

DevinCodes:
@kiyohara the CI is failing due to something unrelated to your changes. No need to worry about that, I'll look into it 😉  I'll try to review this and your other PRs tomorrow, and will for sure take a look at them by the end of the week!

Thank you for your patience 🙇‍♂️

DevinCodes (approved):
Thank you for the PR @kiyohara , this looks great 🙂

kiyohara:
@DevinCodes I'm so glad my PR was accepted.
Thanks for going ahead with the confirmation process :)

PR #474, opened 2022-04-04
fix: #434: confusing empty debug.log generation in gem users root

### LINKED ISSUE #434: Library is creating 'debug.log' file in app directory
- Algolia Client Version: 2.0.1
- Language Version: Ruby

### Description

When we create an `Algolia::Search::Client` instance, the library is creating a `debug.log` file although we have no logging enabled.

### Steps To Reproduce

* Call `Algolia::Search::Client.create(...)` method with valid `app_id` and `api_key` parameters
* Look in your app's root directory to find a `debug.log` file.

kiyohara:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | Fix #434
| Need Doc update   | no

## Describe your change

When you use this gem, your program generate `debug.log` in the working directory. ( ref. #434 )
If you set `ALGOLIA_DEBUG` environment and want to get some algolia gem logging to debug your program, this file is very useful.
However, in almost all situations `debug.log` is a seed of confusion, especially if you are just a program's user that uses this gem.

For general program users, they can't decide why the file is generated because the file is always empty.
Or for program developers who have .gitignore that doesn't contain the line `debug.log`, they might think it's annoying a little.

I understand that the developers can change the behavior by the `:logger` option in your code.
But I'm thinking it's not a bad idea to change the default behavior for the majority of users of this gem.

This PR **don't change** them.
1. If you set `ALGOLIA_DEBUG` env, this gem's behavior is the same as before.
2. If you use methods in this gem with the logger option, this gem's behavior is the same as before.

This PR **will change** below.
1. If you use methods in this gem without the `:logger` option or `ALGOLIA_DEBUG` env, this gem doesn't create `debug.log` in the working directory.
2. If you use `algoliasearch-rails` gem without `ALGOLIA_DEBUG` env as a general use case, this gem doesn't create emptry `debug.log` to your `RAILS_ROOT`.

#434 

---

I'm considering that there is a lot of room for improvement around log handling, for eg method or argument name.
But this PR is focusing on default behavior improvement.
So, even if you feel a name mismatch in this PR code, please miss it.

### DIFF
--- a/lib/algolia/logger_helper.rb
+++ b/lib/algolia/logger_helper.rb
@@ -5,7 +5,7 @@ module Algolia
     # @param debug_file [nil|String] file used to output the logs
     #
     def self.create(debug_file = nil)
-      file              = debug_file || (ENV['ALGOLIA_DEBUG'] ? 'debug.log' : $stderr)
+      file              = debug_file || (ENV['ALGOLIA_DEBUG'] ? File.new('debug.log') : $stderr)
       instance          = ::Logger.new file
       instance.progname = 'algolia'
       instance
--- a/lib/algolia/logger_helper.rb
+++ b/lib/algolia/logger_helper.rb
@@ -5,7 +5,7 @@ module Algolia
     # @param debug_file [nil|String] file used to output the logs
     #
     def self.create(debug_file = nil)
-      file              = debug_file || File.new('debug.log')
+      file              = debug_file || (ENV['ALGOLIA_DEBUG'] ? 'debug.log' : STDERR)
       instance          = ::Logger.new file
       instance.progname = 'algolia'
       instance
--- a/lib/algolia/logger_helper.rb
+++ b/lib/algolia/logger_helper.rb
@@ -5,7 +5,7 @@ module Algolia
     # @param debug_file [nil|String] file used to output the logs
     #
     def self.create(debug_file = nil)
-      file              = debug_file || (ENV['ALGOLIA_DEBUG'] ? 'debug.log' : STDERR)
+      file              = debug_file || (ENV['ALGOLIA_DEBUG'] ? 'debug.log' : $stderr)
       instance          = ::Logger.new file
       instance.progname = 'algolia'
       instance
--- a/lib/algolia/analytics_client.rb
+++ b/lib/algolia/analytics_client.rb
@@ -13,7 +13,7 @@ module Algolia
       def initialize(analytics_config, opts = {})
         @config      = analytics_config
         adapter      = opts[:adapter] || Defaults::ADAPTER
-        logger       = opts[:logger] || LoggerHelper.create('debug.log')
+        logger       = opts[:logger] || LoggerHelper.create
         requester    = opts[:http_requester] || Defaults::REQUESTER_CLASS.new(adapter, logger)
         @transporter = Transport::Transport.new(@config, requester)
       end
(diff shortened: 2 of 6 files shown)


DevinCodes (commented):
Personally I think this change is fine (probably even preferred because I agree it's undesirable to create this log file), but it's a BC-breaking change. Some users might rely on this behavior. I'm going to discuss internally how we should handle this, I'll let you know when we come to a decision 🙂

kiyohara:
I'm very grateful for your consideration!
I'm happy for your and Algolia staff's kindness.

It was really unsettling when I noticed that my little script using the gem was creating `debug.log` in every working directory I run it.
Debug?  What happened? Is malware running?
Once the cause was discovered, it was very simple.
I added the logger overwriting logic to my script.
But it was a little little annoying work.

Hopefully this PR will help more Algolia developers and your team.
Let me know if additional coding is needed!

DevinCodes:
After some discussing internally, I think we may have found a way to make this a non-breaking change, while still preventing the log file from being created if you don't want it. Currently, we only do [logging if the `ALGOLIA_DEBUG` environment variable is set](https://github.com/algolia/algoliasearch-client-ruby/blob/8eb38ebd74cffdcc90c3c19012d68300ebffa614/lib/algolia/http/http_requester.rb#L39) (and has a truthy value). We can change the initializer of the LoggerHelper to still create the `debug.log` file if this environment variable is set, and use `stderr` otherwise as the code snippet in the linked issue metions.

This ensures the gem no longer randomly creates the `debug.log` file, unless you explicitly want to debug. WDYT @kiyohara ?

Do you need me to update the PR to reflect that? Thank you in advance, and thank you for your patience on this 🙇‍♂️

lib/algolia/logger_helper.rb
```diff
@@ -5,7 +5,7 @@ class LoggerHelper
     # @param debug_file [nil|String] file used to output the logs
     #
     def self.create(debug_file = nil)
-      file              = debug_file || File.new('debug.log')
+      file              = debug_file || (ENV['ALGOLIA_DEBUG'] ? 'debug.log' : $stderr)
```
DevinCodes, line 8:
```suggestion
      file              = debug_file || (ENV['ALGOLIA_DEBUG'] ? File.new('debug.log') : $stderr)
```

DevinCodes (changes requested):
Thank you for the quick work on this!! 💯 
I just have one small comment, after that I believe it's good to go 😄

kiyohara:
@DevinCodes Thanks for the suggestion! I have accepted it 😄 
`Logger.new` supports file name string, but I think specifying a File object is more explicit and more resistant to future changes. For example, file permission.
I think this PR has been even better.
May I ask you for a re-review? :pray:

DevinCodes (approved):
Thank you again @kiyohara ! 😄 

We'll release this later this week!

kiyohara:
@DevinCodes I am very happy to help make Algolia's products better.
I'm also very happy that we will be able to upgrade the major version of the gem we use to 2.x.
Thank you so much!

chuckmeyer:
Hi @kiyohara! I'm on the Developer Relations team here at Algolia. @DevinCodes reached out to me because of your recent contributions to our Ruby client and we'd like to send you some thank you swag!

Could you reach out to me over email (chuck.meyer@algolia.com) or Twitter (@chuckm)?  Thank you for your contributions!

PR #478, opened 2022-05-24
fix(multiple_queries): allow params to be nil

joelcogen:
| **Q** | **A** |
|----------------|-------|
| Bug fix?        | yes   |
| New feature?    | no    |
| BC breaks?      | no    |
| Need Doc update | no    |

## Describe your change
Skip call to `to_query_string` in `multiple_queries` if `:params` is nil

`params` is documented as an optional argument of `multiple_queries`, but a recent change introducing a call to `to_query_string` triggers an error if `params` is not present (undefined method `map' for nil:NilClass)

### DIFF
--- a/lib/algolia/search_client.rb
+++ b/lib/algolia/search_client.rb
@@ -479,7 +479,7 @@ module Algolia
       #
       def multiple_queries(queries, opts = {})
         queries.each do |q|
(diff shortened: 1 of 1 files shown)


DevinCodes (approved):
Thank you for the quick fix @joelcogen!

joelcogen:
Fixes #479

PR #492, opened 2023-02-22
chore: move Helpers into Algolia namespace

### LINKED ISSUE #488: `Helpers` module is in global namespace
The `Helpers` module has no namespace currently, which makes collisions very easy to occur. We should namespace this to `Algolia` to avoid these collisions.

DevinCodes:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | yes
| Related Issue     | Fix #488
| Need Doc update   | no

## Describe your change
This moves our `Helpers` module out of the global namespace to prevent possible collisions with other packages or applications. This is technically a BC-breaking change, since some users may rely on this class outside of the code we maintain. This means we likely can't release this until the next major release.

### DIFF
--- a/lib/algolia/helpers.rb
+++ b/lib/algolia/helpers.rb
@@ -1,134 +1,136 @@
 require 'multi_json'
 
-module Helpers
-  # Convert an Hash to json
-  #
-  def to_json(body)
-    body.is_a?(String) ? body : MultiJson.dump(body)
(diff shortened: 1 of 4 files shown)


DevinCodes:
Closing this as it's no longer relevant in the new major version of the API client 🙂

PR #501, opened 2024-07-16
fix(retry): better error

otomatik:
| Q                 | A
| Bug fix?          | yes/no
| New feature?      | yes/no    
| BC breaks?        | no     
| Related Issue     | Fix #...  
| Need Doc update   | yes/no

### DIFF
--- a/test/algolia/unit/retry_strategy_test.rb
+++ b/test/algolia/unit/retry_strategy_test.rb
@@ -86,7 +86,7 @@ class RetryStrategyTest
     end
   end
 
-  describe 'retry stategy decisions' do
+  describe 'retry strategy decisions' do
     def before_all
       super
       @app_id         = 'app_id'
--- a/test/algolia/unit/retry_strategy_test.rb
+++ b/test/algolia/unit/retry_strategy_test.rb
@@ -82,7 +82,7 @@ class RetryStrategyTest
         index.save_object({ objectID: 'one' })
       end
 
-      assert_includes exception.message, 'Unreachable hosts. Last error: SSL_connect'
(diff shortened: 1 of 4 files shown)


millotp (dismissed):
looks good

kombucha (dismissed):
LGTM 👍

kombucha (approved):
Still LGTM

test/algolia/unit/retry_strategy_test.rb
```diff
@@ -82,11 +82,15 @@ def test_failure_when_all_hosts_are_down
         index.save_object({ objectID: 'one' })
       end
 
-      assert_equal 'Unreachable hosts', exception.message
+      assert_includes exception.message, 'Unreachable hosts. Last error for 1.0.0.0: SSL_connect'
+      assert_equal exception.errors, [
+        {:host=>"0.0.0.0", :error=>"SSL_connect SYSCALL returned=5 errno=0 peeraddr=127.0.0.1:443 state=error: certificate verify failed"},
```
DevinCodes, line 87:
```suggestion
        {:host=>"0.0.0.0", :error=>"connection refused: 0.0.0.0:80"},
```

DevinCodes (changes requested):
This should make the CI pass. Without this change, do the tests pass locally?

otomatik:
> This should make the CI pass. Without this change, do the tests pass locally?

yeah it passes locally, maybe it's too specific to the env of test execution then?

DevinCodes:
My guess is that you have a process running on your localhost that accepts connections (but return an SSL error), and for the CI that's not the case. 

Maybe instead of checking the error message exactly, you can check the keys in the hash that's returned instead so we know that an error per host is provided. WDYT?

